PR #16, opened 2017-12-05
Support checking meta

grahamc:


vcunat:
Thanks for finishing this ❤️

PR #34, opened 2018-01-07
factoids: Makes other FAQ entires URL cleaner.

samueldr:
Self-explanatory, the previous anchor links may also have broken easily.


grahamc:
I merged another PR which added a scary URL

Issue #30, opened 2017-12-25
Build a lot more stuff


grahamc:
We've got a good bit of build capacity and it goes largely unused:

![builds](https://user-images.githubusercontent.com/76716/34341778-d0227cf6-e96c-11e7-9fc6-1a7c2d7c92ef.png)

Worse, x86_64-linux has effectively unbounded capacity due to a generous spot instance backing. Darwin is the least strong, and aarch64-linux is in the middle.

For these reasons, I'd like to build way more PRs and use this capacity to make merging PRs easier and safe.

### Phase One: Simple Expansion

 - [ ] Setup sandboxing properly on all the macs
 - [ ] Backport the sandboxing code from master to 17.09
 - [x]  Expand caller access to everyone with merge writes on nixos/nixpkgs

### Phase Two: Easier Build Selection

 - [x] Implement build "sampling" where ofborg automatically selects a handful of attrs to build based on what has changed in the PR.
 - [ ] Implement test sampling where ofborg does the same thing, but selects a handful of changed _tests_.

### Phase Three: Limited Automatic Building

 - [ ] ~Sucker~ ***Convince*** someone to buy me / provide me with a bunch of macs to enroll for capacity
 - [ ] Automatically sample PRs from anyone trusted to call ofborg

### Phase Four: Automatic Building
 - [ ] Automatically sample all PRs

zimbatm:
I asked @macstadium if they could provide some build capacity, no answer so far

7c6f434c:
So, how many macOS builders there are now? Maybe (temporarily) allow to separately ask borg to build just `i686-linux` and `x86_64-linux`, to emphasize abundance of build capacity on this platform?

grahamc:
Right now we have three mac builders. I think that is a good idea, @7c6f434c, and solves the issue around darwin not having adequate sandboxing for this use case. I am planning on the following:

Users bucketed by the following:

 - trusted: explicitly trusted by the ofborg configuration
 - known: has merge rights to nixpkgs
 - unknown: everyone else

1. if you are a trusted user, the build/test job is sent to all architectures.
2. if you are a known user, the build/test job is only sent to architectures which has adequate sandboxing (aarch64-linux, x86_64-linux). 
3. if you are unknown, the command is ignored (as it is now)

With the addition of this change, all known members will be allowed to trigger the eval command.

7c6f434c:
[...]

PR #43, opened 2018-01-27
Allow all Nixpkgs committers to call the bot

grahamc:
Implements https://github.com/NixOS/ofborg/issues/30#issuecomment-360953172 which allows all committers tot Nixpkgs to call the bot and build software on x86_64-linux and aarch64-linux. It also permits all committers to run tests (same rules apply) and also to call the eval command.


grahamc:
In the original version of the description I said it allows committers to build on x86_64-darwin, but I meant aarch64-linux.

PR #58, opened 2018-02-03
Automatically schedule builds for trusted users

grahamc:
cc #30


7c6f434c:
Does it filter out `[WIP]`-tagged PRs?

grahamc:
Not right now, but that might be a good issue to raise. If we add exceptions to when it doesn't trigger builds, we'll need to provide more feedback about what it is doing and why.

grahamc:
It now filters out WIPs.

PR #59, opened 2018-02-03
Auto tag topic issue 56

### LINKED ISSUE #56: Add tags based on files that are touched.
If certain Python-related files are touched I would like to have a `python` tag.

I don't think we could use the codeowners file, but we may use a similar file. Basically, a list of regexes and a tag.

grahamc:
Closes #56

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -40,6 +40,18 @@
             "zimbatm"
         ]
     },
+    "tag_paths": {
+        "6.topic: python": [
+            "pkgs/top-level/python-packages.nix",
+            "pkgs/development/interpreters/python",
+            "pkgs/development/python-modules",
+            "doc/languages-frameworks/python.md"
+        ],
+        "6.topic: ruby": [
+            "pkgs/development/interpreters/ruby",
+            "pkgs/development/ruby-modules"
+        ]
+    },
     "checkout": {
         "root": "/var/lib/gc-of-borg/.nix-test-rs"
     },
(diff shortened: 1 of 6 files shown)


config.public.json
```diff
@@ -40,6 +40,18 @@
...
+        "6.topic: python": [
+            "pkgs/top-level/python-packages.nix",
+            "pkgs/development/interpreters/python",
+            "pkgs/development/python-modules",
+            "doc/languages-frameworks/python.md"
+        ],
+        "6.topic: ruby": [
+            "pkgs/development/interpreters/ruby",
+            "pkgs/development/ruby-modules"
+        ]
```
grahamc, line 53: cc @peti, @zimbatm (because I just assumed you'd like this... I can take it out ;)), @NixOS/darwin-maintainers @gleber (for erlang) @ttuegel (for kde)

The way this auto-tagger works is:

1. a PR is created
2. it touches a file described by this list and the tag is added
3. a commit is pushed which makes it no longer touch a file described by this list
4. the tag is removed

In other words, this behavior would also override a human's judgement.

Would any of you like to add to this?

gleber: @grahamc Sorry don't have time to send a PR, but here's a list for Erlang / Elixir / BEAM:

```
doc/languages-frameworks/beam.xml
pkgs/development/beam-modules/
pkgs/development/interpreters/elixir/
pkgs/development/interpreters/erlang/
pkgs/development/tools/build-managers/rebar/
pkgs/development/tools/build-managers/rebar3/
pkgs/development/tools/erlang/
pkgs/development/tools/elixir/
pkgs/top-level/beam-packages.nix
```

grahamc: Done in 5f12e4b29c12ecf9d7735f7135f37f667a08bb2c

PR #69, opened 2018-02-09
Adds a `message_type` field to `MessageFromIRC`.

samueldr:
The `message_type` allows switching to `notice` type messages.

It defaults to `privmsg` and any other (string) values will have the
same effect than using `privmsg`.

No code should need to be updated after this change.

* * * 

### Why ?

This is because of the switch to using `{^_^}` to send channel advance messages. The switch is great, but (imho) channel notices were much easier to tone out from conversations than normal privmsg. This is because of the way my IRC client defaults to presenting those messages. The same can be said of how botbot presents those messages.

I also personally believe that those *scheduled* messages, and not responses from user actions, should be sent as notices instead of messages.

Also:

```
[19:47:53] <gchristensen>  here is what it would take:
[19:48:33] <gchristensen> https://github.com/NixOS/ofborg/blob/released/ircbot/src/bin/gateway.rs#L30 <- add an Option<MessageType> where MessageType is an enum { Privmsg, Notice }
[19:48:53] <gchristensen> https://github.com/NixOS/ofborg/blob/released/ircbot/src/bin/gateway.rs#L82-L86
[19:49:04] <gchristensen> actually, don't make it an Option :)
[19:49:13] <gchristensen> using the Privmsg vs. Notice field here, to determine if it sends a privmsg or a notice
[19:50:01] <gchristensen> then https://github.com/grahamc/network/blob/master/zoidberg/nix-channel-monitor/changes.sh#L16 where it says \"target\" also specify \"message_type\" (or whatever): "notice"
[...]
[19:56:34] <samueldr> the hard part is running the code to see what I'm doing :)
[19:56:44] <gchristensen> hmm
[19:56:48] <gchristensen> that will be hard
[19:57:04] <gchristensen> but... `cargo build` should provide sufficient puking to tell you if it is correct
[19:57:09] <gchristensen> not ideal though :(
```

This is why this is untested code. It does compile though! See [network#4](https://github.com/grahamc/network/pull/4) for the other part.

### DIFF
--- a/ircbot/src/bin/gateway.rs
+++ b/ircbot/src/bin/gateway.rs
@@ -30,7 +30,9 @@ use std::env;
 #[derive(Serialize, Deserialize, Debug)]
 struct MessageToIRC {
     target: String,
-    body: String
+    body: String,
+    #[serde(default = "default_irc_message_type")]
+    message_type: String
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -38,6 +40,10 @@ struct MessageFromIRC {
     from: String,
     sender: String,
     body: String
+}
+
+fn default_irc_message_type() -> String {
+    "privmsg".to_string()
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


ircbot/src/bin/gateway.rs
```diff
@@ -30,7 +30,9 @@ use std::env;
 #[derive(Serialize, Deserialize, Debug)]
 struct MessageToIRC {
     target: String,
-    body: String
+    body: String,
+    #[serde(default = "default_irc_message_type")]
```
grahamc, line 34: Nice.

ircbot/src/bin/gateway.rs
```diff
@@ -30,7 +30,9 @@ use std::env;
...
+    message_type: String
```
grahamc: This would be a bit better as an `enum`, for example: https://play.rust-lang.org/?gist=786e78e1678ea7b22ebc6c5fab5d9b82&version=stable

Do you mind giving that a try?

samueldr:
If you want this squashed, I'll be glad to squash it.

I think I did it okay for the Enum, starting from your example.

PR #77, opened 2018-02-15
add nix operation type

LnL7:
moves some of the nix command implementation to the nix module and makes it a bit easier to test.


grahamc:
Awesome, thanks!

PR #78, opened 2018-02-16
fix stdenvs test if nixos system channel does not exist

LnL7:
Ask nix to resolve `<nixpkgs>` instead of hardcoding it.


ofborg/src/tasks/massrebuilder.rs
```diff
@@ -740,14 +740,24 @@ fn parse_commit_messages(messages: Vec<String>) -> Vec<String> {
 mod tests {
 
     use super::*;
+    use std::process::Command;
 
     #[test]
     fn stdenv_checking() {
+        let output = Command::new("nix-instantiate")
```
LnL7, line 747: Is there some way to do this at configure/build time and use a constant instead?

PR #87, opened 2018-02-24
Evaluation filter

grahamc:
Move evaluation filtering in to Rust.


grahamc:
Closes #26 and reduces the number of lines of PHP (including dependencies) by 50,000.

PR #91, opened 2018-02-26
fix test with ANSI escape characters

LnL7:
The nix commands print the error: commands in red, the ANSI reset caused
one of the tests to fail.


LnL7:
```
test result: ok. 61 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

😃

grahamc:
OMG nice!

Issue #17, opened 2017-12-08
Be able to run oborg's meta checks on a local branch


grahamc:
Be able to do like:

```
$ cd nixpkgs
$ ofborg eval-checks

grahamcofborg-eval-nixos-manual — LGTM
grahamcofborg-eval-nixos-options — LGTM
grahamcofborg-eval-nixpkgs-manual — LGTM
grahamcofborg-eval-nixpkgs-tarball — LGTM
grahamcofborg-eval-nixpkgs-unstable-jobset — LGTM
grahamcofborg-eval-package-list — LGTM

Would label: 10.rebuild-linux:1-10
Would label: 10.rebuild-darwin:0
```

wolfgangwalther:
meta checks are now implemented in nixpkgs' GHA. Most of them can be run locally with `nix-build ci` (eval is run separately). Those that are not implemented, yet, are tracked in https://github.com/NixOS/nixpkgs/issues/355847.

PR #98, opened 2018-03-06
Add me to config.extra-known-users.json

oxij:
=^.^=

### DIFF
--- a/config.extra-known-users.json
+++ b/config.extra-known-users.json
@@ -2,5 +2,6 @@
     "bhipple",
     "dywedir",
     "edef1c",
+    "oxij",
     "unode"
 ]


grahamc:
Ok you're able to call it now. Please read through almost all of the readme, you can stop at "running a builder": https://github.com/NixOS/ofborg

oxij:
Ah, "Running meta checks locally" is new and useful. Thanks.

Would also be nice to have a local command to see what exactly would get rebuilt against master (i.e. what ofborg uses for tags).

grahamc:
Indeed: https://github.com/NixOS/ofborg/issues/17

PR #103, opened 2018-03-09
add xeji to extra-known-users

xeji:
as suggested by @Mic92 in https://github.com/NixOS/nixpkgs/pull/36676#issuecomment-371885984 - thank you!


grahamc:
Deployed!

PR #105, opened 2018-03-10
Add abbradar to trusted_users

abbradar:
I want to fix several packages on Darwin but don't have a test machine -- having our trusty bot's help would be nice.


grahamc:
Deployed!

Issue #90, opened 2018-02-26
Spurious failures on unsupported systems


shlevy:
If I touch a package that is known not to build on darwin, I get an error like https://github.com/NixOS/nixpkgs/pull/35666#issuecomment-368476092

7c6f434c:
Timeouts could also benefit from being reported separately. Unlike platform checks, timeout failures can sometimes require fetching the whole log (if more than ten packages depend on the long-build one)

7c6f434c:
I am not sure how feasible it is to recognise out-of-space failures separately…

PR #106, opened 2018-03-10
Don't try and build stuff that cannot be built

grahamc:


xeji:
really like that new feature, thank you!

grahamc:
Closes #90

PR #108, opened 2018-03-11
Record build status - 0.1.3

grahamc:
Send BuildResult messages to the log queue for collection. Hint: we may want to make the github comment poster no longer listen on the build-results queue, but instead of the log queue. This is a bit complicated, though, since the current design of the build-result queue makes it much more likely to only post one comment per result. Note this doesn't SAVE the results of the build, but gets the builders sending the data.

### DIFF
--- a/ofborg/src/tasks/log_message_collector.rs
+++ b/ofborg/src/tasks/log_message_collector.rs
@@ -10,6 +10,7 @@ use std::io::Write;
 
 use ofborg::writetoline::LineWriter;
 use ofborg::message::buildlogmsg::{BuildLogStart, BuildLogMsg};
+use ofborg::message::buildresult::BuildResult;
 use ofborg::worker;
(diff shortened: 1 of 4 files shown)


ofborg/src/tasks/log_message_collector.rs
```diff
@@ -158,20 +165,26 @@ impl worker::SimpleWorker for LogMessageCollector {
...
+                message = MsgType::Start(msg);
             } else {
-                return Err(format!("failed to decode job: {:?}", decode_msg));
+                let decode_msg: Result<BuildResult, _> = serde_json::from_slice(body);
+                if let Ok(msg) = decode_msg {
+                    attempt_id = msg.attempt_id.clone();
+                    message = MsgType::Finish(msg);
+                } else {
+                    return Err(format!("failed to decode job: {:?}", decode_msg));
+                }
```
grahamc, line 187: This is ugly, huh?

PR #110, opened 2018-03-11
add va1entin to extra known users

va1entin:
Hey @grahamc 
I recently became maintainer for pynacl in nixpkgs (https://github.com/nixos/nixpkgs/blob/master/pkgs/development/python-modules/pynacl/default.nix) and wanted to kindly ask for permission to use your awesome build automation with this


grahamc:
You got it!

PR #119, opened 2018-03-14
Update config.known-users.json

gilligan:
Add "gilligan" to the list of known users


grahamc:
Ah, this is generated. Please put it in the extra-known-users file

PR #134, opened 2018-03-18
Log instantiation failures

grahamc:


grahamc:
Bug: the snippet log isn't being sent when no attempt is made.

ofborg/src/tasks/build.rs
```diff
@@ -94,6 +96,10 @@ impl<'a, 'b> JobActions<'a, 'b> {
         };
     }
 
+    pub fn log_snippet(&self) -> VecDeque<String> {
+        self.snippet_log.clone()
```
Ekleog: What about returning a `&VecDeque<String>` (with `&self.snippet_log`), so that the caller can `clone()` if need be?

ofborg/src/tasks/build.rs
```diff
@@ -55,6 +55,7 @@ pub struct JobActions<'a, 'b> {
     receiver: &'a mut notifyworker::NotificationReceiver,
     job: &'b buildjob::BuildJob,
     line_counter: u64,
+    snippet_log: VecDeque<String>,
```
Ekleog, line 58: A ring buffer would likely make more sense, both from a performance and a code simplicity standpoint, but maybe the added dependency isn't worth it?

ofborg/src/tasks/build.rs
```diff
@@ -364,10 +376,10 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
 
         println!("ok built ({:?}), building", success);
         println!("Lines:\n-----8<-----");
-        snippet_log.iter().inspect(|x| println!("{}", x)).last();
+        actions.log_snippet().iter().inspect(|x| println!("{}", x)).last();
         println!("----->8-----");
 
-        let last10lines: Vec<String> = snippet_log.into_iter().collect::<Vec<String>>();
+        let last10lines: Vec<String> = actions.log_snippet().into_iter().collect::<Vec<String>>();
```
Ekleog: If following my above-mentioned advice of returning an `&VecDeque<_>` from `log_snippet()`, this should just become `.iter()`, so that the returned value isn't consumed. Also, the `::<Vec<String>>` isn't required if there's the `: Vec<String>` type annotation on the variable declaration, only one is needed for type inference to work, usually (and here I'd guess none is actually needed, as the `last10lines` variable is then passed to a function, thus forcing its type).

ofborg/src/nix.rs
```diff
@@ -102,16 +104,35 @@ impl Nix {
...
+                     Err(f) => Err((attr.clone(), lines_from_file(f)))
+                 }
+            )
+            .collect();
+
+        let (ok, err): (
+            Vec<Result<String, (String, Vec<String>)>>,
+            Vec<Result<String, (String, Vec<String>)>>)  = attr_instantiations
+            .into_iter()
+            .partition(|x| x.is_ok());
```
Ekleog: I think this could be shortened (simplified?) into `let (ok, err): (Vec<_>, Vec<_>) = attr_instantiations.into_iter().partition(|x| x.is_ok());`
EDIT: actually, see below, this may not be actually a useful point to make if this part is to be deleted anyway.

ofborg/src/nix.rs
```diff
@@ -102,16 +104,35 @@ impl Nix {
...
+
+        let ok_ret: Vec<String> = ok.into_iter().map(|x| x.unwrap()).collect();
+        let err_ret: Vec<(String, Vec<String>)> = err.into_iter().map(|x| x.unwrap_err()).collect();
+
+        return (
+            ok_ret,
+            err_ret
+        )
```
Ekleog: Here I think it'd be cleaner (and likely much faster, though that's not the point) to just not try using iterators, and have `let mut ok_ret; let mut err_ret;`, do a big `for` look on the `attrs`, and just `push` to both vectors.

Something like (untested & typed in the github comment box):
```rust
let mut ok = Vec::new();
let mut err = Vec::new();
for attr in attr_instantiations.into_iter() {
    match self.safely_instantiate_attrs(nixpkgs, file, vec![attr.clone()]) {
        Ok(_) => { ok.push(attr); }
        Err(f) => { err.push((attr, lines_from_file(f))); }
    }
}
(ok, err)
```
(also, it's standard practice to omit the `return` keyword where possible)

Ekleog (commented):
Mostly style points, as I'm not used enough with ofborg's codebase (yet?) to make actually useful points :+1:

ofborg/src/tasks/build.rs
grahamc: I had actually already changed this to return a Vec<String> to be easier to pass to the build-result message.

ofborg/src/tasks/build.rs
grahamc: Yeah, a ring buffer would be nice here... perhaps an easy thing for a newbie to contribute (cc @LnL7)

LnL7: I'm spending most of my time on ZHF right now, but that sounds like a fun distraction 😄

LnL7: https://doc.rust-lang.org/std/collections/vec_deque/
`This queue has O(1) amortized inserts and removals from both ends of the container.`

Ekleog: Yeah, the main reason for wanting a ring buffer is it means it'd allow to avoid having to manually maintain the invariant by hand, not performance (which, if I remember correctly how deques are implemented, should still be better in the constant, thanks to not allocating/deallocating memory frequently)

Ekleog: (eg. https://github.com/NixOS/ofborg/pull/134/files/cfa9547434ea9b7f8666f6f140fc342f73741da9#diff-d9ced247c4766cbb09e7fe716d230b20R172 for the manually-maintaining-the-invariant point)

PR #135, opened 2018-03-18
Simplify log_snippet

Ekleog:


grahamc:
Magic..!

Ekleog:
Yeah, type inference guesses the return type of `.into()` is to be `Vec<_>` thanks to the return type of the function, thus calls `VecDeque::into<Vec<_>>`, which calls `Vec::from<VecDeque<_>>`, which does the work, if I remember correctly how the `From` and `Into` traits are implemented :)

PR #151, opened 2018-03-31
Add infinisil to config.extra-known-users

infinisil:
Would like to be more helpful and faster at reviewing the ever increasing amount of PR's :D


grahamc:
I'll update this ticket when I deploy it.

grahamc:
Deployed!

infinisil:
Thanks!

PR #159, opened 2018-04-11
add green checkmarks

LnL7:
https://travis-ci.org/LnL7/ofborg/builds

### DIFF
--- a/ofborg/test-srcs/make-pr.sh
+++ b/ofborg/test-srcs/make-pr.sh
@@ -4,31 +4,21 @@ set -eu
 bare=$1
 co=$2
 
-(
+makepr() {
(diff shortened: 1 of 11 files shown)


grahamc:
:heart_eyes_cat:

ofborg/test-srcs/make-pr.sh
```diff
@@ -1,36 +1,24 @@
 #!/bin/sh
```
LnL7: Maybe this should be `#!/usr/bin/env bash` to avoid other POSIX sh issues in the future.

shlevy: :100:

PR #161, opened 2018-04-12
Add timokau to trusted users

timokau:
I'm opening a lot of PRs recently and would love to be able to check those changes against more platforms than `x86_64-linux`. "trusted" might be optimistic, but testing on darwin would be nice so I thought I'd try.

I'll gladly open another PR for "known" instead of "trusted" if you think thats better for now.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -54,6 +54,7 @@
             "ryantm",
             "shlevy",
             "srhb",
+            "timokau",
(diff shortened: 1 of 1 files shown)


7c6f434c (approved):

grahamc:

timokau:
Thanks!

PR #165, opened 2018-04-26
Add gilligan to trusted users

gilligan:


grahamc:

PR #166, opened 2018-04-27
Asking for ofborg access

jlesquembre:
As suggested here: https://github.com/NixOS/nixpkgs/pull/39568#issuecomment-384896678


grahamc:

PR #167, opened 2018-04-27
add woffs to config.extra-known-users

woffs:
- as suggested by @Mic92 in https://github.com/NixOS/nixpkgs/pull/39569
- and sort ;-)


grahamc:

PR #171, opened 2018-05-04
Export NIX_REMOTE in the shell so that building on recent NixOS doesn't fail

Ekleog:
I guess this is “fixing” the symptoms and not the cause, but…?


Ekleog:
Superseded by #172, which does the same thing but better.

PR #175, opened 2018-05-15
Add yurrriq to config.extra-known-users

yurrriq:
Thanks for this project! I'd love to get added to the extra-known-users.


yurrriq:
Thanks!

PR #176, opened 2018-05-21
abort mass-rebuilder if less than 8Gb of memory is "available"

obadz:
…where available is defined as "how much RAM can the kernel give to an application immediately without having to reach for swap".

@grahamc, are you sure we shouldn't instead check for total memory assuming other sources of memory usage can be pushed to swap?

Coded up at the London Rust meetup https://www.meetup.com/Rust-London-User-Group/events/250178008/


obadz:
I assume I need to regenerate the carnix file but that does not seem to work?

```
$ sh nix/update-carnix.sh 
error: Found argument './../ofborg/Cargo.lock' which wasn't expected, or isn't valid in this context
```

```
$ carnix --version
carnix 0.8.5
```

PR #190, opened 2018-07-07
Adds samueldr to config.extra-known-users.json

samueldr:
I hope to help with nixpkgs PRs.


grahamc:
Maybe youd like to be a trusted user?

PR #191, opened 2018-07-07
add hydra links to GitHub comments

ryantm:
If a build is attempted, this tries to construct a valid link to the Hydra builds for the attrs. Sometimes it will fail because Hydra won't have a job for it.

### DIFF
--- a/ofborg/src/tasks/githubcommentposter.rs
+++ b/ofborg/src/tasks/githubcommentposter.rs
@@ -99,7 +99,25 @@ fn result_to_comment(result: &BuildResult) -> String {
     reply.push("".to_owned());
 
     if let Some(ref attempted) = result.attempted_attrs {
-        reply.extend(list_segment("Attempted", attempted.clone()));
+        let jobset =
+            match result.pr.target_branch {
+                None => "trunk",
+                Some(ref s) =>
+                    match s.as_ref() {
+                        "staging" => "staging",
+                        _ => "trunk",
+                    },
+            };
+        reply.extend(list_segment(
+            "Attempted",
+            attempted.iter().map( |attempt| {
+                format!(
+                    "[{}](https://hydra.nixos.org/job/nixpkgs/{}/{}.{})",
+                    attempt,
+                    jobset,
+                    attempt,
+                    result.system)
+            }).collect()));
     }
 
     if let Some(ref skipped) = result.skipped_attrs {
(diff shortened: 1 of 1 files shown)


ryantm:
The test that is failing on Travis `tests::log_message_collector::test_logs_collect` is passing on my machine.

LnL7:
Maybe there's a race condition in the test, can't  reproduce it either and the second build succeeded.

ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -99,7 +99,17 @@ fn result_to_comment(result: &BuildResult) -> String {
     reply.push("".to_owned());
 
     if let Some(ref attempted) = result.attempted_attrs {
-        reply.extend(list_segment("Attempted", attempted.clone()));
+        let target_branch = result.pr.clone().target_branch.unwrap_or("master".to_string());
+        reply.extend(list_segment(
+            "Attempted",
+            attempted.iter().map( |attempt| {
+                format!(
+                    "[{}](https://hydra.nixos.org/job/nixpkgs/{}/{}.{})",
```
LnL7, line 115: Using the branch isn't a good idea, none of the jobsets have the same name except for staging.

ryantm:
@LnL7 Okay, I think I fixed the issue of branch to jobset correspondence.

ryantm:
```
07:53 < ryantm> gchristensen: What do you think of https://github.com/NixOS/ofborg/pull/191 ?
07:53 < {^_^}> #191 (by ryantm, open): add hydra links to GitHub comments
07:55 < gchristensen> I don't like it much, because of the chances of 404s
07:55 < gchristensen> it is possible to be a not very strong signal
07:55 < gchristensen> I think it would be nicer to build before and after the commit
08:01 < gchristensen> because, then, too, much better signal can be provided: specifics on how the closure size has changed, if it could build before / could build now, build time
                      differences, build closure changes, etc
08:04 < ryantm> gchristensen: You're saying even if it doesn't 404 it might not be a good signal, right?
08:04 < gchristensen> yeah
08:04 < gchristensen> because then you have to know what you're looking for and how to find it, and then to know what it means
08:07 < ryantm> gchristensen: well, building before and after sounds great, and build caching might mean it's often pretty quick.
08:08 < ryantm> gchristensen: I'll go ahead and close that PR.
08:08 < gchristensen> true
08:08 < gchristensen> maybe we can fetch data from the Hydra API (or add data to the Hydra API so it is useful)
```

PR #192, opened 2018-07-07
Add uri-canva as trusted user.

uri-canva:
So I can run Darwin builds on hydra, for example #43018 and #42832.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -60,6 +60,7 @@
             "shlevy",
             "srhb",
             "timokau",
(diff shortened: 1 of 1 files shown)


grahamc:
I think it is too soon to add you as a trusted user, sorry. I'd be okay with an extra-known-user (search PRs for an example)

uri-canva:
Sounds good, will update this PR.

grahamc:
Closing until you do update it.

PR #196, opened 2018-07-13
Add Profpatsch to trusted users

Profpatsch:


config.public.json
```diff
@@ -51,6 +51,7 @@
             "orivej",
             "peti",
             "primeos",
+            "Profpatsch",
```
grahamc, line 54: Needs to be lowercase, I'll fix it.

PR #205, opened 2018-08-03
Add basvandijk to known users

basvandijk:
I would like to run the bot on my PRs.

### DIFF
--- a/config.known-users.json
+++ b/config.known-users.json
@@ -17,6 +17,7 @@
       "armijnhemel",
       "aszlig",
       "aycanirican",
+      "basvandijk",
       "bendlas",
       "benley",
       "bennofs",


infinisil:
Apparently this list is autogenerated from people in the NixOS organization by scripts/update-known-users.sh, running it should then put you in that list (and all other new people in it, such as me).

extra-known-users.json seems to be used for people not in the organization, by manually updating it

LnL7:
Yeah, NixOS members are automatically allowed. Add yourself to trusted users if you'd also like to be able to run darwin build.

basvandijk:
Ok thanks for the info. I can't run ` scripts/update-known-users.sh` myself since [it requires access to `config.private.json`](https://github.com/NixOS/ofborg/blob/released/scripts/update-known-users.sh#L4) which I don't have. So I'll just wait until somebody else runs it.

I will make a PR to add myself to trusted users because I maintain a few packages that should build on darwin.

PR #206, opened 2018-08-05
detect timedout builds

LnL7:
This should change the error messages for timed out builds to something like.

```
Timed out, unknown build status on x86_64-linux
```

I think this is fully backwards compatible, allowing builders to continue sending build results in the old format, while newer versions send more information using the new format. This could be cleaned up once all the infrastructure is updated.

### DIFF
--- a/ofborg/src/asynccmd.rs
+++ b/ofborg/src/asynccmd.rs
@@ -3,9 +3,10 @@ use std::thread;
 use std::collections::HashMap;
 use std::process::Stdio;
 use std::process::ExitStatus;
-use std::sync::mpsc::sync_channel;
 use std::process::Command;
 use std::io::Read;
+use std::sync::mpsc;
+use std::sync::mpsc::sync_channel;
 use std::sync::mpsc::{SyncSender, Receiver};
 use std::io::BufReader;
 use std::io::BufRead;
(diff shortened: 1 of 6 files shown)


ofborg/src/asynccmd.rs
```diff
@@ -234,9 +238,9 @@ mod tests {
         let acmd = AsyncCmd::new(cmd);
 
         let mut spawned = acmd.spawn();
-        let lines: Vec<String> = spawned.lines().into_iter().collect();
+        let lines: Vec<String> = spawned.lines().collect();
         assert_eq!(lines, vec!["stdout", "stderr", "stdout2", "stderr2"]);
-        let ret = spawned.wait().unwrap().unwrap().unwrap().success();
```
grahamc, line 239: :heart_eyes:

ofborg/src/tasks/build.rs
```diff
@@ -364,35 +370,40 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+                Some(101) => BuildStatus::TimedOut, // nix build timedout
+                Some(i) => BuildStatus::UnexpectedError {
+                    err: format!("command failed with exit code {}", i),
+                },
+                None => BuildStatus::UnexpectedError {
+                    err: "unexpected build failure".into(),
+                },
+            },
+            e => BuildStatus::UnexpectedError {
+                err: format!("failed on interior command {:?}", e),
```
LnL7, line 397: I don't think it's a problem in practice, but this could potentially expose more information about the builder then before.

ofborg/src/asynccmd.rs
```diff
@@ -188,12 +189,15 @@ impl AsyncCmd {
...
-        &self.rx
+    pub fn lines<'a>(&'a mut self) -> mpsc::Iter<'a, String> {
+        self.rx.iter()
     }
 
-    pub fn wait(self) -> thread::Result<Option<Result<ExitStatus, io::Error>>> {
+    pub fn wait(self) -> Result<ExitStatus, io::Error> {
         self.waiter.join()
+            .map_err(|_err| io::Error::new(io::ErrorKind::Other, "Couldn't join thread."))
+            .and_then(|opt| opt.ok_or(io::Error::new(io::ErrorKind::Other, "Thread didn't return an exit status.")))
```
LnL7, line 199: I'm not entirely sure these messages are correct.

PR #208, opened 2018-08-07
Add myself to trusted_users

basvandijk:
I maintain a few packages that should build on darwin and would like to use the bot to build and test these packages.


basvandijk:
Thank you.

PR #211, opened 2018-08-07
0.1.7 for the new result changes

grahamc:
This should fail PR checks... so, testing that!

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -23,9 +23,6 @@ in {
   in pkgs.runCommand "ofborg-rs-symlink-compat" {
     src = stripDeps build;
   } ''
-
-    set -x
-
     mkdir -p $out/bin
     for f in $(find $src -type f); do
       bn=$(basename "$f")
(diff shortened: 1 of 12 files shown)


default.nix
```diff
@@ -43,6 +40,32 @@ in {
     test -e $out/bin/github_comment_poster
     test -e $out/bin/log_message_collector
     test -e $out/bin/evaluation_filter
+
+    # Verify that the outpath contains the version number matching the
```
LnL7, line 44: nice 😃

grahamc:
@LnL7 can you take a look to see why the tests are failing now? Seems related to your timeout PR.

ofborg/test-srcs/build-pr/default.nix
```diff
@@ -4,7 +4,7 @@ in {
   success = derivation {
     name = "success";
     system = builtins.currentSystem;
-    builder = nix.shell;
+    builder = builtins.storePath nix.shell;
```
LnL7, line 7: still works on darwin with build-remote.

LnL7 (approved):

PR #213, opened 2018-08-08
Package up dependencies with composer2nix

grahamc:
I had to patch composer2nix with:

require_once(dirname(__FILE__)./../../../../vendor/autoload.php);

Loads of stuff to check in, but oh well.


grahamc:
A flaky test is making this fail  checks, merging anyway.

PR #215, opened 2018-08-09
Tagged comments

grahamc:
I think it is safe to merge this, but better wait until @Globin updates their ofborg to 0.1.7.

The request ID is generated as part of the comment parser job, so all the results from one comment request will share a REQUEST_ID.


ofborg/src/message/buildresult.rs
```diff
@@ -17,7 +17,7 @@ pub struct BuildResult {
     pub system: String,
     pub output: Vec<String>,
     pub attempt_id: String,
-    pub request_id: Option<String>,
```
LnL7, line 20: This was introduced some time around the log viewer, right?

grahamc: Yeah, I think so!

PR #218, opened 2018-08-10
buildresult: replace serialization with an enum to simplify compatibility

LnL7:
This separates the serialized BuildResult type that requires backwards
compatibility to avoid breaking results received from older builders and
the LegacyBuildResult type use internally.


ofborg/src/message/buildresult.rs
```diff
@@ -10,16 +10,103 @@ pub enum BuildStatus {
...
+}
+
+impl BuildResult {
+    pub fn legacy(&self) -> LegacyBuildResult {
+        // TODO: replace this with simpler structs for specific usecases, since
+        // it's decouples the structs from serialization.  These can be changed
+        // as long as we can translate all enum variants.
+        match self {
+            BuildResult::Legacy { repo, pr, system, output, attempt_id, request_id, attempted_attrs, skipped_attrs, .. } =>
+                LegacyBuildResult {
```
LnL7: I can also clean this part up a bit if you prefer, didn't want to make a huge pr.

ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -541,69 +522,6 @@ foo
 The following builds were skipped because they don't evaluate on x86_64-linux: not-attempted
 
 No partial log is available.
-"
-        );
-    }
-
-    #[test]
-    pub fn test_no_status_compatibility() {
```
LnL7, line 556: This is tested by the the serialization now.

LnL7:
I added a V1 format with tagging and cleaned up the usage a bit.

PR #219, opened 2018-08-14
Add kalbasit to trusted users

kalbasit:
I was trying to review https://github.com/NixOS/nixpkgs/pull/45006 and I quickly realized that I can't build on Darwin, the target of this PR.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -40,6 +40,7 @@
             "jb55",
             "joachifm",
             "jtojnar",
+            "kalbasit",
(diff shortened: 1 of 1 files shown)


LnL7:
Sorry, because of sandboxing limitations we have to limit darwin builds for now. Hopefully we can fix this and allow more people in the future.

LnL7:
It's been a while and and he's a member now. /cc @grahamc

PR #221, opened 2018-08-15
Add myself to trusted users

Enzime:
I've made a few PRs fixing builds on Darwin and I hope to make a few more, so it'd be really great if I could be able to use the bot to verify my builds on macOS as well as Linux simultaneously.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -29,6 +29,7 @@
             "dotlambda",
             "dtzwill",
             "edolstra",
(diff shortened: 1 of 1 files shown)


grahamc:
Sorry, I can't merge this right now: darwin access has to be carefully controlled right now.

If you open a PR adding you to the extra-known-users list I'll merge that.

PR #224, opened 2018-08-18
Move infinisil to trusted_users

infinisil:
Would like to help with getting packages to build properly on Darwin :)

Btw: I almost committed the file with using tabs instead of spaces. Would be nice to add an [`.editorconfig`](https://editorconfig.org/) file, such that editors (with the plugin installed) take care of such things automatically. nixpkgs also uses this

### DIFF
--- a/config.extra-known-users.json
+++ b/config.extra-known-users.json
@@ -9,7 +9,6 @@
     "etu",
     "fgaz",
     "grwlf",
-    "infinisil",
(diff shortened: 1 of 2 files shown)


LnL7:
Deployed.

I like the idea of an editorconfig, for single files I usually add a modeline but that's editor specific and isn't even possible in json.

PR #239, opened 2018-09-18
Add worldofpeace to extra-known-users

worldofpeace:
@xeji suggested that I do this.


worldofpeace:
Huh, I was expecting a special initiation @grahamc :rofl:

PR #250, opened 2018-10-14
add new eval checker for the package list with allowAliases = false

### LINKED ISSUE #248: feature request: check evaluation succeeds with `allowAliases = false`
People are often using aliases to provide backward compatibility, but as a general rule, we should not allow aliases to be used inside of nixpkgs, because it will break evaluation when the configuration option `allowAliases = false` is used.

It would be nice if ofborg could check evaluation with this configuration option to detect these alias additions that do not update all the old sites in nixpkgs.

ryantm:
closes #248

By checking the package list without aliases, we can make it more
likely that people do not introduce new aliases into `nixpkgs` without
removing them from the repository. Aliases are supposed to be for
backward compatibility.

Also, we can make sure this configuration option continues to work for
people. Noteably, this configuration option is used by
r-ryantm/nixpkgs-update to avoid submitting duplicate pull requests
for the package and its alias, and to use the correct (non-alaised)
name in the pull request.

### DIFF
--- a/ofborg/src/tasks/massrebuilder.rs
+++ b/ofborg/src/tasks/massrebuilder.rs
@@ -332,6 +332,19 @@ impl<E: stats::SysEvents + 'static> worker::SimpleWorker for MassRebuildWorker<E
             ),
 
             EvalChecker::new(
+                "package-list-no-aliases",
+                nix::Operation::QueryPackagesJSON,
+                vec![
+                    String::from("--file"),
+                    String::from("."),
+                    String::from("--arg"),
+                    String::from("config"),
+                    String::from("{ allowAliases = false; }"),
+                ],
+                self.nix.clone()
+            ),
+
+            EvalChecker::new(
                 "nixos-options",
                 nix::Operation::Instantiate,
                 vec![


ofborg/src/tasks/massrebuilder.rs
```diff
@@ -331,6 +331,19 @@ impl<E: stats::SysEvents + 'static> worker::SimpleWorker for MassRebuildWorker<E
                 self.nix.clone()
             ),
 
+            EvalChecker::new(
+                "package-list-no-aliases)",
```
7c6f434c: Looks like parenthesis mismatch.

ryantm:
@7c6f434c fixed! Thanks for reviewing.

7c6f434c (approved):

LnL7:
@grahamc Do you think this would be ok in terms of resource usage, given that it adds an extra evaluation to each merge event?

grahamc:
I think yes, let's see how it goes. We can change the architecture of this process if we need to. This is something we should check, since it prevents problems from reaching Hydra.

PR #251, opened 2018-10-15
Add florianjacob to extra-known-users.json

florianjacob:
Is this the right thing to do to be able to run commands like `@GrahamcOfBorg test matrix-synapse` by myself as e.g. in https://github.com/NixOS/nixpkgs/pull/41728#issuecomment-395954947 ? :bowing_man:


florianjacob:
Thanks! :smiley:

PR #253, opened 2018-10-16
Add c0bw3b to extra-known-users

c0bw3b:
I promise I won't abuse OfBorg resources :)

### DIFF
--- a/config.extra-known-users.json
+++ b/config.extra-known-users.json
@@ -3,6 +3,7 @@
     "bhipple",
     "bignaux",
     "brainrape",
+    "c0bw3b",
     "costrouc",
     "dywedir",
     "edef1c",


samueldr:
[Aren't you already there?](https://github.com/NixOS/ofborg/blob/85a3b3341e0455903e21d0d84af91c525136888d/config.known-users.json#L27)

Maybe you intended to add yourself to [trusted users](https://github.com/NixOS/ofborg/blob/78285dc4a14643ea1bc91e614ead54478be71b86/config.public.json#L14)?

c0bw3b:
Oh I am already? _Noice_!
I never asked before so no I just aimed for "known user" for starter but didn't even check beforehand...

I'll close this and go with it for know. I don't have any Darwin env to test on my own and don't really want to use OfBorg as a Darwin crash-tester. :)

PR #254, opened 2018-10-17
Add alyssais to extra-known-users

alyssais:
See 
https://github.com/NixOS/nixpkgs/pull/48629#issuecomment-430680221.

cc @zimbatm.


grahamc:
Thank you!

PR #255, opened 2018-10-21
move Ma27/markuskowa/dywedir to trusted users

Mic92:
Both are nixpkgs maintainer now.


Mic92:
cc @Ma27 @markuskowa

Ma27 (approved):

Ma27 (approved):

Mic92:
cc @dywedir

markuskowa (approved):

dywedir:
Thanks!

grahamc:
Note not all nixpkgs contributors are in the trusted user list. After discussing with @LnL7, we both agree these three seem good.

Ma27:
> After discussing with @LnL7, we both agree these three seem good.

I'm taking that as a compliment ;)

But seriously, thanks a lot for trusting me so much by adding me as nixpkgs maintainer and listing me as trusted borg user!

PR #258, opened 2018-10-23
Prohibit use of nixpkgs.extend from inside Nixpkgs

roberth:
`extend` is a new function on Nixpkgs that lets users extend Nixpkgs with overlays after calling the Nixpkgs function. This makes Nixpkgs similar to package sets that were created with `lib.makeExtensible`.

Nixpkgs should probably not start using this, because re-evaluating the Nixpkgs is somewhat expensive.

See https://github.com/NixOS/nixpkgs/pull/47430#issuecomment-432254467

I am not familiar with ofborg and I wasn't sure how to test this, so it is mostly cargo culted. I did test the ad-hoc overlay itself.

### DIFF
--- a/ofborg/src/tasks/massrebuilder.rs
+++ b/ofborg/src/tasks/massrebuilder.rs
@@ -345,6 +345,19 @@ impl<E: stats::SysEvents + 'static> worker::SimpleWorker for MassRebuildWorker<E
             ),
 
             EvalChecker::new(
+                "package-list-no-extend",
+                nix::Operation::QueryPackagesJSON,
+                vec![
+                    String::from("--file"),
+                    String::from("."),
+                    String::from("--arg"),
+                    String::from("overlays"),
+                    String::from("[(self: super: { extend = abort ''The pkgs.extend function should not be used inside Nixpkgs. Extensive use decreases performance and complicates the package set. If absolutely necessary, this restriction may be lifted.''; })]"),
+                ],
+                self.nix.clone()
+            ),
+
+            EvalChecker::new(
                 "nixos-options",
                 nix::Operation::Instantiate,
                 vec![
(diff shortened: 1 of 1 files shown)


LnL7:
I'm not sure if it makes sense to run a separate eval for each feature like this. I can't really think of any issues with just running a single strict eval to check for issues.

roberth:
@LnL7 better?

LnL7:
Yeah, but maybe this should be a config option in nixpkgs. That way it's a bit less obscure to reproduce.

roberth:
@LnL7 I don't want people to avoid `extend` when they actually need it, so, to be honest, I think this check is not for users, and therefore I don't think it should be an option. If someone really wants this check in their own tools, it's not too complicated to reproduce.

If someone is motivated to make it an option, I can review their PR.

roberth:
cc @Ericson2314

LnL7:
Having the error message inline in the command just feels a bit weird to me, but yeah it's not that complicated.

@grahamc What do you think?

Ericson2314 (approved):
I like config option too, but this is fine for now. I've asked enough things from @roberth with this :).

grahamc:
I'm sorry to be annoying and not provide more actionable feedback sooner.

This test should be applied to the evaluation which happens before and after a PR is merged. Please move the configuration to https://github.com/NixOS/ofborg/blob/ae982a2a683dfb5062ef8fb1b03bb7c3804b7e31/ofborg/src/outpaths.nix and delete the changes to the massrebuilder.rs.

Aliases are tested this way because aliases are desirable for external users, but can easily rot unless they are checked regularly. This is why the check is performed once with, and once without aliases.

I don't believe `extends` will ever rot like aliases can, so no need to perform the different checks.

roberth:
I'm cleaning up my open PR list. After almost four years, this has proven itself to be unimportant in many ways. If anyone disagrees, feel free to modify and use the code in any way you like.

PR #271, opened 2018-11-17
Move c0bw3b to trusted users

c0bw3b:
Politely asking :smiley_cat: 
it would allow me to check Linux+Darwin package updates before merging into nixpkgs.

I understood that Darwin builds are not sandboxed properly so trusted users should be extra careful when invoking ofBorg.
I also keep in mind that builders have limited resources and that I should check https://monitoring.nix.ci before launching a potentially heavy job.

/cc @grahamc @LnL7 @Mic92

### DIFF
--- a/config.known-users.json
+++ b/config.known-users.json
@@ -24,7 +24,6 @@
       "bjornfor",
       "bluescreen303",
       "brainrape",
-      "c0bw3b",
       "chaoflow",
       "cillianderoiste",
       "copumpkin",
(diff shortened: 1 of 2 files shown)


7c6f434c:
>I understood that Darwin builds are not sandboxed properly so trusted users should be extra careful when invoking ofBorg.

… and opening a PR not marked as WIP might also invoke ofBorg without an explicit request.

</pointless-nitpick>

This is just a nitpick about the wording of the list of caveats, in case it ever gets linked from outside as an example of a list of caveats.

Mic92 (approved):

c0bw3b:
Thanks Graham

PR #274, opened 2018-11-23
README.md: switch to workaround for running tests

Mic92:


Mic92:
Btw. It would be also cool to have tests that are affected by changes in the package list. Then `nix-review` could pick them up.

PR #276, opened 2018-11-23
massrebuilder.rs: correct the shape of nixpkgs argument to release.nix

7c6f434c:
To allow restricted mode evaluation, the Nixpkgs path has to be passed on the command line; this PR fixes the shape of the argument (it needs to have the actual path as `outPath` and have a couple extra entries).

### DIFF
--- a/ofborg/src/tasks/massrebuilder.rs
+++ b/ofborg/src/tasks/massrebuilder.rs
@@ -350,7 +350,7 @@ impl<E: stats::SysEvents + 'static> worker::SimpleWorker for MassRebuildWorker<E
                 vec![
                     String::from("--arg"),
                     String::from("nixpkgs"),
-                    String::from("./."),
+                    String::from("{ outPath=./.; revCount=999999; shortRev=\"ofborg\"; }"),
                     String::from("./nixos/release.nix"),
                     String::from("-A"),
                     String::from("options"),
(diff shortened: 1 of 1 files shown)


grahamc:
I think this test will need updating: https://github.com/NixOS/ofborg/pull/189/files#diff-c5ddc8807759886d253ec47844ada69fR9 -- you can run locally with `cargo test`

7c6f434c:
I think `cargo test` just succeeds for me with that test as-is

grahamc:
```18:24 <MichaelRaskin> Trying to run cargo test… it takes time to download all the dependencies
18:27 <MichaelRaskin> Hm. I think cargo test succeeds even without fixing things…
18:28 <gchristensen> hrm
18:28 <gchristensen> :)
18:28 <MichaelRaskin> I might be doing something wrong, but cargo test says
18:29 <MichaelRaskin> test nix::tests::instantiation_nixpkgs_restricted_mode ... ok
18:29 <gchristensen> hrm :/
18:29 <gchristensen> oh
18:29 <gchristensen> because the test doesn't care about the attrset, it just tostrings it
18:29 <gchristensen> which is fine
18:29 <MichaelRaskin> It tests a different simpler thing, indeed
18:29 <gchristensen> ok
18:29 <gchristensen> well
18:29 <gchristensen> let's get this out there
```

PR #282, opened 2018-12-01
Checks

grahamc:


tilpner:
image

The tab labels are all identical, which makes them useless. They could instead display the difference, here x86_64-darwin vs aarch64-linux

7c6f434c:
Drawback: now I won't get any notifications when the build is finished…

PR #289, opened 2018-12-10
Add non-bogus completion times to GitHub checks

edef1c:
This should eventually be replaced with completion times from the builders, but they don't currently provide that information, so local time will have to do.

### DIFF
--- a/ofborg/src/tasks/githubcommentposter.rs
+++ b/ofborg/src/tasks/githubcommentposter.rs
@@ -8,6 +8,7 @@ use hubcaps::checks::{CheckRunOptions, Output, Conclusion, CheckRunState};
 use ofborg::message::buildresult::{BuildStatus, BuildResult, LegacyBuildResult};
 use ofborg::worker;
 use amqp::protocol::basic::{Deliver, BasicProperties};
+use chrono::{DateTime, Utc};
 
 
 pub struct GitHubCommentPoster {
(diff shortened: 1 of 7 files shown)


ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -658,13 +660,15 @@ No partial log is available.
...
+
         assert_eq!(
-            result_to_check(&result),
+            result_to_check(&result, timestamp),
             CheckRunOptions {
                 name: "nix-build -A bar -A foo --argstr system x86_64-linux".to_string(),
                 actions: None,
                 started_at: None,
-                completed_at: Some("2018-01-01T01:01:01Z".to_string()),
+                completed_at: Some("2018-04-20T13:37:42Z".to_string()),
```
grahamc: :eyes:

grahamc (approved):
lgtm, waiting for checks

ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -138,7 +139,7 @@ fn result_to_check(result: &LegacyBuildResult) -> CheckRunOptions {
             result.system
         ),
         actions: None,
-        completed_at: Some("2018-01-01T01:01:01Z".to_string()),
+        completed_at: Some(timestamp.to_rfc3339()),
```
grahamc: Seems chrono isn't emitting `Z` but instead `completed_at: Some("2018-04-20T13:37:42+00:00")` and thus each of these tests are failing. We'd need to use (and probably should?) https://docs.rs/chrono/0.4.6/chrono/struct.DateTime.html#method.to_rfc3339_opts with use_z set to True for it to retain the same format, like this:

```
assert_eq!(dt.to_rfc3339_opts(SecondsFormat::Secs, true),
           "2018-01-26T18:30:09Z");
```

grahamc: Note I specifically prefer using the `Z` format and not changing the output format, since we know that specific format works and not changing the output here makes it a much safer change. Despite iso8601 being a "standard" who knows how much of it they follow.

grahamc:
Nice! Thank you!

PR #290, opened 2018-12-14
Move @hedning to trusted users

hedning:
Help prevent me breaking darwin builds :)


grahamc:
I'll deploy this in about 12hrs.

PR #293, opened 2019-01-01
Experiment/maintainer ping

grahamc:
Ping maintainers based on the algorithm:

1. identify all the files changed by a PR
2. identify all the attributes changed by a PR
3. identify the file which defines each changed attribute
4. filter out all attributes whose containing file is not modified by the PR
5. ping those attributes' maintainers

Presently it is "no-op": posting the list of packages under the grahamcofborg-eval-check-maintainers check to a gist and not pinging anyone. After a few days of "burning in" and seeing how it goes, I plan on actually pinging people.

Here is a demo: https://github.com/NixOS/nixpkgs/pull/53142

output: https://gist.github.com/GrahamcOfBorg/fa9a84d85a7ad2633df025f2735dfa22


tilpner:
Does it need to say `pkgs.` for every single entry?

ofborg/src/maintainers.rs
```diff
@@ -0,0 +1,172 @@
+use std::collections::HashMap;
+use ofborg::nix::Nix;
+use std::path::Path;
+use tempfile::NamedTempFile;
+use std::io::Write;
```
tilpner: What's your import ordering system? It looks either random or append-to-bottom

ofborg/src/maintainers.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+impl From<std::string::FromUtf8Error> for CalculationError {
+    fn from(e: std::string::FromUtf8Error) -> CalculationError {
+        CalculationError::Utf8(e)
+    }
+}
+
+impl ImpactedMaintainers {
+    pub fn calculate(nix: &Nix, checkout: &Path, paths: &Vec<String>, attributes: &
+                     Vec<Vec<&str>>) -> Result<ImpactedMaintainers, CalculationError> {
```
tilpner: `&Vec` is never what you want, use `&[...]`

grahamc:
No, removed in 8c2a87a. Thanks!

ofborg/src/maintainers.rs
```diff
@@ -0,0 +1,172 @@
...
+            .checkout_origin_ref(&OsStr::new("master"))
+            .unwrap();
+
+        let paths = working_co.files_changed_from_head(&hash).unwrap();
+
+        working_co
+            .checkout_ref(&OsStr::new(&hash))
+            .unwrap();
+
+        let remote = env::var("NIX_REMOTE").unwrap_or("".to_owned());
```
tilpner, line 158: It's very subjective, but I prefer `String::new()`. I haven't checked whether `"".to_owned()` gets away with not heap allocations, but `String::new()` does. Feel free to ignore

ofborg/src/nix.rs
```diff
@@ -150,7 +161,26 @@ impl Nix {
...
+        let mut attrargs: Vec<String> = Vec::with_capacity(2 + (argstrs.len() * 3));
+        attrargs.push("--expr".to_owned());
+        attrargs.push(expr.to_owned());
+        for (argname, argstr) in argstrs {
+            attrargs.push(String::from("--argstr"));
+            attrargs.push(argname.to_owned());
+            attrargs.push(argstr.to_owned());
+        }
+
+        return self.safe_command(Operation::Evaluate, nixpkgs, attrargs, extra_paths);
```
tilpner: Unnecessary return

ofborg/src/nix.rs
```diff
@@ -150,7 +161,26 @@ impl Nix {
             attrargs.push(attr);
         }
 
-        return self.safe_command(Operation::Instantiate, nixpkgs, attrargs);
+        return self.safe_command(Operation::Instantiate, nixpkgs, attrargs, vec![]);
```
tilpner: Unnecessary return

ofborg/src/maintainers.nix
```diff
@@ -0,0 +1,110 @@
+{ changedattrsjson, changedpathsjson }:
+let
+  pkgs = import ./. {};
+
```
FRidh, line 4: maybe a `with pkgs.lib; with builtins;` to improve the readability. Namespacing is typically good, but I don't think it's worth it here. Or do you want to keep track of what's provided by `lib` since it's vendored?

ofborg/src/maintainers.nix
```diff
@@ -0,0 +1,110 @@
...
+  changedattrs = builtins.fromJSON (builtins.readFile changedattrsjson);
+  changedpaths = builtins.fromJSON (builtins.readFile changedpathsjson);
+
+  anyMatchingFile = filename:
+    let
+      matching = builtins.filter
+        (changed: pkgs.lib.strings.hasSuffix changed filename)
+        changedpaths;
+    in (builtins.length matching) > 0;
```
FRidh, line 13: ( `isEmpty = list: (length list) == 0` function in `lib` would be nice :) )

ofborg/src/maintainers.nix
```diff
@@ -0,0 +1,110 @@
...
+    (builtins.length (builtins.filter anyMatchingFile files)) > 0;
+
+  enrichedAttrs = builtins.map
+    (path: {
+      path = path;
+      name = builtins.concatStringsSep "." path;
+    })
+    changedattrs;
+
+validPackageAttributes = builtins.filter
```
FRidh: indentation changes

ofborg/src/maintainers.nix
```diff
@@ -0,0 +1,110 @@
...
+  (pkg:
+    if (pkgs.lib.attrsets.hasAttrByPath pkg.path pkgs) then
+      (if (builtins.tryEval (pkgs.lib.attrsets.attrByPath pkg.path null pkgs)).success
+       then true
+       else builtins.trace "Failed to access ${pkg.name} even though it exists" false)
+    else builtins.trace "Failed to locate ${pkg.name}." false
+  )
+  enrichedAttrs;
+
+  attrsWithPackages = builtins.map
```
FRidh, line 35: indentation changes

FRidh (commented):
The algorithm looks good to me.

7c6f434c:
>3. identify the file which defines each changed attribute
>4. filter out all attributes whose containing file is not modified by the PR

What does it mean «defines the changed attribute»?

I have a feeling that either we will have problems with all-packages.nix, or with Firefox/Wine (where bumps go in one file and mkDerivation goes into another).

Were both these cases considered, or should I try to work out the specifics?

(I assume that a change also includes addition/removal; by the way, how do we want to handle aliases.nix ?)

FRidh:
@7c6f434c those cases the algorithm indeed won't cache.

As a work-around, we may continue on 4) with:
- if an attribute doesn't have a transitive parent that is already in the result of 4), then extend 4) with that one. I think a breadth first search is needed for this.

It won't solve it entirely, but I don't think we can do any better than that.

7c6f434c:
>@7c6f434c those cases the algorithm indeed won't cache.

There is a moderate chance of false-positive on something interesting in all-packages.nix or aliases.

>As a work-around, we may continue on 4) with:
>- if an attribute doesn't have a transitive parent that is already in the result of 4), then extend 4) with that one. I think a breadth first search is needed for this.

I think if you do that (and even if you just don't special-case large files), it is also a good idea to have logic like «going to ping more than 6 people → file an issue in ofborg instead» (you can just ping me, I will file the issue manually)

grahamc:
The current algorithm says an attribute is changed if the derivation it points to changed.

ofborg/src/maintainers.nix
grahamc: Wishing for a nice nixfmt :)

grahamc:
I've rebased this on top of my branch where I rustfmt'd and cargo clippy'd the codebase, and formatted the Nix code with the changes in nix-mode at https://github.com/NixOS/nix-mode/pull/61. I'm thinking I'll probably deploy this in the morning unless someone would like to review it more carefully.

grahamc:
For people having questions about how ofborg handles' all-packages.nix: https://github.com/NixOS/nixpkgs/pull/53281

it didn't ping anyone because the change was only in all-packages.nix.

7c6f434c:
OK, so we do need a combination of `all-packages.nix` defined packages, and a mass rebuild, to trigger weird behaviour. Although indeed `src` for an override defined in `all-packages.nix` is unfortunate.

PR #294, opened 2019-01-02
Add puffnfresh to trusted users

puffnfresh:
As suggested by @ryantm


grahamc:
I'll deploy this in about 12hrs.

PR #301, opened 2019-01-14
WIP: Git config name and email

Profpatsch:
git requires an `user.name` and `user.email` to be set for most actions. ofborg shouldn’t depend on the system’s `--global` settings, instead it should provide config options.

This PR is not yet doing that for `clone.rs`:
https://github.com/NixOS/ofborg/blob/released/ofborg/src/clone.rs#L64-L69

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -76,6 +76,7 @@ let
     name = "gh-event-forwarder";
     buildInputs = with pkgs; [
       bash
+      nix
       nix-prefetch-git
       rust.rustc
       rust.cargo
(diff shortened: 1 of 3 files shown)


Profpatsch:
Setting

```
GIT_AUTHOR_NAME
GIT_AUTHOR_EMAIL
```

for each git operation ofborg does should be the way to go (see `man git-commit-tree`).

shell.nix
```diff
@@ -76,6 +76,7 @@ let
     name = "gh-event-forwarder";
     buildInputs = with pkgs; [
       bash
+      nix
```
JohnAZoidberg, line 79: Why does this PR need to add nix?

Profpatsch: See commit description of the corresponding commit: https://github.com/NixOS/ofborg/pull/301/commits/5d73c771f0dae5df0bcaab7f9e4cbe9b2766ae87

Artturin:
Can be closed, fixed by https://github.com/nixos/ofborg/commit/e64d033fb6de15fa0cfc30964ab93c435d442a8b

PR #309, opened 2019-01-26
Fixup tests darwin

grahamc:
Since oborg creates "queued" stubs now, we can't send a job for macos only for it to be dropped on the floor later :)


grahamc:
See https://github.com/NixOS/nixpkgs/pull/54574/checks for an example of why this is bad.

PR #320, opened 2019-02-12
add teto to trusted maintainers

teto:
I've been working on kernel and lua infrastructure recently and would like to trigger darwin builds especially without annoying my fellow @7c6f434c  https://github.com/NixOS/nixpkgs/pull/55305


7c6f434c:
I do like the work performed on Lua, and the easier it is for @teto to check on Darwin, the fewer things break there!

PR #331, opened 2019-03-17
extra-known-users: Add myself (lilyball)

lilyball:
Also add myself to `known-users`.

I hope I'm doing this right. It wasn't clear from the README what the process was, but it looks like `known-users` is generated by a script.

### DIFF
--- a/config.known-users.json
+++ b/config.known-users.json
@@ -96,6 +96,7 @@
       "knedlsepp",
       "kosmikus",
       "lheckemann",
+      "lilyball",
       "lnl7",
       "lovek323",
       "lschuermann",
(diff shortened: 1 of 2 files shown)


config.known-users.json
```diff
@@ -96,6 +96,7 @@
       "knedlsepp",
       "kosmikus",
       "lheckemann",
+      "lilyball",
```
grahamc, line 99: Adding yourself to both lists is not necessary, just extra-known-users is sufficient. This one will be regenerated with out you in it, so it is fine.

lilyball: I figured it would be regenerated but I didn't know how frequently that happens.

PR #334, opened 2019-03-21
known-users: add illegalprime

illegalprime:
I'm interested in getting some cross-compilation PRs through, it was suggested to me to trigger builds myself by adding myself to ofborg, these are the PRs I have, a few more should be coming soon:
https://github.com/NixOS/nixpkgs/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Aillegalprime+

I hope this is right, and thank you!

### DIFF
--- a/config.known-users.json
+++ b/config.known-users.json
@@ -76,6 +76,7 @@
       "grwlf",
       "hedning",
       "hrdinka",
+      "illegalprime",
       "imalsogreg",
       "infinisil",
       "izorkin",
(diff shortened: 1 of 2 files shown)


config.known-users.json
```diff
@@ -76,6 +76,7 @@
       "grwlf",
       "hedning",
       "hrdinka",
+      "illegalprime",
```
grahamc, line 79: You're not a known user, this list is generated from people who have merge rights. Anyway, it doesn't matter -- this list gets regenerated and you'll be removed. The entry in the other list is sufficient.

PR #337, opened 2019-03-22
config.extra-known-users.json: Add myself (balsoft)

balsoft:
Adds "balsoft" to the list of extra known users.


grahamc:
Thank you!

PR #341, opened 2019-03-24
outpaths.nix: Makes callable

samueldr:
The behaviour of the callable outpaths.nix is such that it will call
nix-env as expected to run the equivalent ofborg step, in the CWD.

The directory it is ran on can be changed using `--argstr path ...`.


samueldr:
Of note: `~/tmp/ofborg/ofborg/src/outpaths.nix --show-trace` works as expected :)

PR #347, opened 2019-04-11
Add lua topic paths

teto:
interpreter/packages etc.

Ideally, it would be cool if someone could manually add the "lua" topic to my ~5 already existing lua PRs (since I don't think ofborg will look at older ones): https://github.com/NixOS/nixpkgs/pulls/teto


teto:
thanks

PR #349, opened 2019-04-12
GitHub token vending machine

grahamc:
Fetch API Creds based on the user / repo needed, instead of hard coding IDs in the config.


grahamc:
Need to verify  this caches clients and the client doesn't have to fetch new tokens on each request.

PR #362, opened 2019-05-08
Use more widely supported arrow symbol in performance report

### LINKED ISSUE #360: Use less obscure arrow symbols in performance report
The issue, as seen on my machine

Currently, [the `🡖` character (U+1F856, "South East Sans-Serif Arrow")](https://www.compart.com/en/unicode/U+1F856) is used in performance reports. This character doesn't seem to be very widely supported by fonts yet, as it was only introduced in 2014, with Unicode 7.0. 

I propose changing this to [the `↘` (U+2198, "South East Arrow")](https://www.compart.com/en/unicode/U+2198), which has much wider support, since it has been in the standard since 1993, with Unicode 1.1. 

This has the effect of rendering the character as an emoji on some platforms, which doesn't seem like a huge downside, possibly an upside.

Synthetica9:
Fixes #360.


Synthetica9:
Just noticed that that test doesn't include any "up" arrows, that might be worth testing

PR #388, opened 2019-08-21
move @jonringer to trusted users

mmahut:
I would like to move @jonringer to trusted users.

Jon is helping with PR requests and going a great job, giving him the ability to build on darwin would save a lot of (also cpu) time. :)


jonringer:
Thanks for the recommendation :) :+1:

PR #395, opened 2019-09-03
add label for exactly 1 rebuild

ryantm:
It is fundamentally different to review a leaf package with no other reverse dependencies in Nixpkgs and one that has at least one reverse dependency, so make those separate labels.

I expect rolling this out is going to have to coincide with updating the labels and adding the new labels.

### DIFF
--- a/ofborg/src/notifyworker.rs
+++ b/ofborg/src/notifyworker.rs
@@ -11,7 +11,7 @@ pub struct NotifyWorker<T: SimpleNotifyWorker> {
 pub trait SimpleNotifyWorker {
     type J;
 
-    fn consumer(&self, job: &Self::J, notifier: &mut NotificationReceiver);
+    fn consumer(&self, job: &Self::J, notifier: &mut dyn NotificationReceiver);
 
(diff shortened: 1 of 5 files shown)


grahamc:
This is a good idea. I'd like to continue applying the `1-10` label in the `1` case, so existing searches continue to work.

ryantm:
@grahamc Okay, done, and I added the proper tests now. Also, your suggestion makes this change will be backward compatible, so that's nice!

By the way, I didn't commit this but in order to build this I had to add the `dyn` annotation in 7 places because of this error:

`error: trait objects without an explicit `dyn` are deprecated`

If you want me to add those `dyn` annotations to this branch, I can.

grahamc:
Please do add the annotations :)

ryantm:
@grahamc Added.

ryantm:
@grahamc Do you need anything else from me to merge this?

ryantm:
Even though the CI is failing, I think this is ready to go, because the CI errors seem unrelated.

PR #408, opened 2019-10-25
d-goldin: Adding myself

d-goldin:
I'd like to be able to test some PRs a bit better myself without
relying on another contributor for everything.

### DIFF
--- a/config.extra-known-users.json
+++ b/config.extra-known-users.json
@@ -17,6 +17,7 @@
     "danieldk",
     "delroth",
     "devhell",
+    "d-goldin",
     "Ekleog",
     "ElvishJerricco",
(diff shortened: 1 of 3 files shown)


d-goldin:
Ideally I'd also like to be able to build on darwin, especially because I have zero access to any darwin resources myself, but I'm not sure what's required for that level of trust.

d-goldin:
@fpletz: As discussed. Thanks in advance.

fpletz:
I am happy to vouch for @d-goldin since his technical expertise is great, he has already contributed to nixpkgs in the past and he wants to contribute even more.

d-goldin:
Thanks!

PR #419, opened 2019-11-17
add @Kiwi to trusted users

Kiwi:
I'm trying to get more active contributing to nixpkgs so making myself known to ofborg seems like a good idea and also I'd like to add support for macOS to a few packages I already maintain sometime real soon (tm) so trusted users it is.

I'm like really trustworthy, I've had access to the community aarch64 server for months and only broke it once! Just look at my face :smirk: (jk, I haven't broken it... yet) Oh, and also @grahamc said to, so there's that. :smile:

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -55,6 +55,7 @@
             "jonringer",
             "jtojnar",
             "kalbasit",
+            "kiwi",
             "knedlsepp",
             "lassulus",
             "lheckemann",


config.public.json
```diff
@@ -55,6 +55,7 @@
             "jonringer",
             "jtojnar",
             "kalbasit",
+	    "kiwi",
```
ryantm: Seems like your indentation is different.

ryantm (changes requested):

config.public.json
Kiwi: Somehow a \t snuck in... I never use tabs. Weird.

ryantm (approved):

worldofpeace (approved):

worldofpeace:
:eyes:

Kiwi:
It only took me a little over a year but I now have a macOS builder and have been updating and testing as much as I can on aarch64-linux, x86_64-linux, and x86_64-darwin! :smile: I still don't know how to use the bot. Anyway, I should have them all done about the same time as nixpkgs supports the new apple arm machines or at least soon after... so I'll check back in same time next year and let you know how it's going. :smiley:

PR #421, opened 2019-11-22
maintainers: Add support for teams

jtojnar:
* meta.maintainers can contain either individual maintainers or maintainer teams
* maintainer teams can be nested
* multiple teams can have overlapping membership


wolfgangwalther:
This needs to be done in nixpkgs now (and I think was partly/mostly done already?).

PR #425, opened 2019-12-15
config.public.json: add marsam to trusted_users

marsam:
I'd appreciate if I could test my changes on Darwin


worldofpeace (approved):
:exclamation: And the fact you contribute prominently to darwin in nixpkgs makes this seem strange it's not already granted. :+1:

PR #436, opened 2020-02-21
Update to current nixpkgs-unstable

andir:
This required the introduction of an overlay file since ofBorg still
requires openssl 1.0.2. Updating to newer version of openssl also
requires updating the amqp library which requires some more involved
changes.

The php code is only updated to php 7.2 and not 7.3 (the default in
nixpkgs) since there seem to been a few syntax changes that aren't yet
compatible with composer2nix.

In previous versions of the rust infrastructure within nixpkgs we would
have been able to override all crates during the invocation of the
target crate. That `override` feature was removed as it caused a huge
eval overhead for larger projects. We did end up with (n^2)
instantiations of dependencies since they were being overridden on every
invocation of every dependency on every level further down the chain.
The current understanding is that the build tooling that each project is
using (e.g.  crate2nix, crate2nix, …) that is driving the actual build
could easily reintroduce that feature without the overhead.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-{ pkgs ? import ./nix {} }:
+{ pkgs ? import ./nix { overlays = [ (import ./nix/overlay.nix) ]; } }:
 
 let
   ofborgOverrides = {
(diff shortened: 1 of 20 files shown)


andir:
I am still trying to figure out what it takes to make the PHP stuff work again.. Right now it is just spewing out cryptic error messages about `break` and `continue 2`.

andir:
The default php version of unstable is 7.3 which did break composer2nix. For now we can probably just stick with 7.2.

andir:
Ok, CI finally passed. It looks good. Was a bit more involved due to the rust version upgrades and the corresponding linting fixes.

PR #438, opened 2020-03-15
Add check for lib tests

infinisil:
Without this, a big part of the lib tests aren't being done, which
previously lead to e.g. https://github.com/NixOS/nixpkgs/pull/76861
And this will also be useful for checked maintainers in https://github.com/NixOS/nixpkgs/pull/82461

I have not tested this code in any way


andir:
This looks really useful. @grahamc who is able to review/merge things here? Would be nice to see our CI evolve into checking more tings like outlined here.

PR #439, opened 2020-03-16
add symphorien to known users

symphorien:


Ekleog:
:+1:

veprbl:
I was just about to submit the same request.

grahamc:

veprbl:
@grahamc So everyone can build on darwin or not? I think I saw a PR without a darwin build that day.

veprbl:
I guess should target trusted_users in config.public.json instead

cole-h:
Yes, being added to `trusted_users` in config.public.json is necessary for Darwin builds, while (I believe) everybody can ask ofborg to build `aarch64-linux` and `x86_64-linux`.

PR #442, opened 2020-03-29
ofborg: pass rev

flokli:
Hydra seems to have passed in rev all the time, and nixpkgs will soon
require `rev` to be set.
    
This initially broke nixpkgs eval in
https://github.com/NixOS/nixpkgs/commit/5e8545e72341887bb371407a71a723bc0e9c7844,
so we now update ofborg to always pass `rev` when instantiating nixpkgs
to avoid breaking it.

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -406,6 +406,12 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
     }
 
     fn evaluation_checks(&self) -> Vec<EvalChecker> {
+        // the value that's passed as the nixpkgs arg
+        let nixpkgs_arg_value = format!(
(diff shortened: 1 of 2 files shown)


ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -406,6 +406,12 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
     }
 
     fn evaluation_checks(&self) -> Vec<EvalChecker> {
+        // the value that's passed as the nixpkgs arg
+        let nixpkgs_arg_value = format!(
+            "{{ outPath=./.; revCount=999999; shortRev=\"{}\"; rev=\"{}\"; }}",
+            &self.job.pr.head_sha[0..7],
+            &self.job.pr.head_sha,
+        );
```
grahamc, line 414: wow you're getting fancy here with the head_sha stuff :P

flokli: Well, at least here it's easy to get it, so we can very well pass in the proper value.

PR #444, opened 2020-04-01
2018-ification

cole-h:
2018-ification changes include:
* replacing `extern crate` imports (except for ones annotated with
`#[macro_use]` to prevent touching more files; maybe in a follow-up)
* removing standalone imports; for example, `use serde_json;` -- the
code already uses `serde_json::*` where necessary

Minor formatting changes include:
* collapsing imports from the same root (e.g. `use std::io::Read; use
std::io::BufRead` -> `use std::io::{BufRead, Read};`
* separating `crate` imports from external imports from `std` imports --
this is how many project seem to order their imports, most notably the
official Rust repo
* moving some `use`s from the first line to below the file's doc comment
* changing file-level doc comments from `///` to `//!` (so they actually
show up as documentation)
* `rustfmt` messing with the `nom` macros in `src/commentparser.rs`

Most of the 2018-ification was done by `cargo fix --edition-idioms`, but
all changes were manually inspected to catch any detrimental changes,
and `cargo test`/`cargo check` still succeed.

---

Verified code impact (none) with `cd ofborg && nix-shell ../shell.nix --run 'cargo check && cargo test'`.

Supercedes https://github.com/NixOS/ofborg/pull/417.

Small note: when I ran `cargo fix --edition-idioms`, it introduced anonymous lifetimes (`<'_>`) to a few structs. I undid this change for the sole reason that I am still unfamiliar with lifetimes. `rust_2018_idioms` is the lint that notices this -- reason being: `hidden lifetime parameters in types are deprecated`. If this is desirable, I can add it back in (currently sitting in a `git stash` entry).


grahamc:
This is awesome, thank you so much!

Issue #446, opened 2020-04-03
Some PRs trigger: code: 2, kind: NotFound, message: "No such file or directory"


doronbehar:
I see [this error](https://gist.github.com/GrahamcOfBorg/709b8b2b85b5d64ff6bdc2f25d0f8f87) on several PRs with seemingly OK syntax:

```
Failed during the setup of executing nix-env: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Examples:

- https://github.com/NixOS/nixpkgs/pull/84004
- https://github.com/NixOS/nixpkgs/pull/84150
- https://github.com/NixOS/nixpkgs/pull/84160

If the PRs themselves have something wrong with them, I think at least the error message could be better?

doronbehar:
There's also [this error](https://github.com/NixOS/nixpkgs/pull/84183#event-3196110875) coming out for me.

cole-h:
This is because ofborg currently propagates IO errors with little-to-no context. The problematic function could be one of these two:

https://github.com/NixOS/ofborg/blob/a87225519e6bd349f67180f2d862f600d437f420/ofborg/src/nixenv.rs#L62
https://github.com/NixOS/ofborg/blob/a87225519e6bd349f67180f2d862f600d437f420/ofborg/src/nixenv.rs#L69

because any `std::io::Error` gets transformed to the local `Error` enum via this line:

https://github.com/NixOS/ofborg/blob/a87225519e6bd349f67180f2d862f600d437f420/ofborg/src/nixenv.rs#L124

losing all context. I'm working on a PR for this that should give more context, like what operation actually failed (was it creating the file? Removing it? Writing to it?), but I don't know what other information would be helpful, or even if that other information could be displayed.

---

The internal error tag is when something goes really wrong inside ofborg. Currently, Graham manually removes these (in order to allow himself to investigate the cause, first). Assuming it evals + builds, I would say don't worry about it (unless someone else says otherwise).

cole-h:
The root cause appears to be: if the `nix` invocation fails (like when evaluating), the stats file won't be created (example log: https://gist.github.com/GrahamcOfBorg/9dfb034692a8b8397c6bdc476c6910ec). My follow-up PR at #449 demotes this to a logged warning, and should let it get to the point where ofborg reports an actual evaluation error.

EDIT: If this resurfaces, feel free to re-open or open a new issue and ping me. I'll look into it.

PR #447, opened 2020-04-03
nixenv: describe what went wrong when io fails

cole-h:
Instead of just propagating a raw `std::io::Error`, wrap that in our own
`Error` enum to provide more context for what actually went wrong.

---

Should provide more information for situations like https://github.com/NixOS/ofborg/issues/446. Underlying cause still needs to be investigated (hopefully made easier by this change).

### DIFF
--- a/ofborg/src/nixenv.rs
+++ b/ofborg/src/nixenv.rs
@@ -61,29 +61,18 @@ impl HydraNixEnv {
     /// emulates Hydra's behavior.
     fn place_nix(&self) -> Result<(), Error> {
         let outpath = self.outpath_nix_path();
+        let mut file = File::create(&outpath).map_err(|e| Error::CreateFile(outpath, e))?;
 
-        let mut file = match File::create(&outpath) {
-            Ok(f) => f,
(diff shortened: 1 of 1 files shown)


ofborg/src/nixenv.rs
```diff
@@ -59,16 +59,32 @@ impl HydraNixEnv {
...
     /// Put outpaths.nix in to the project root, which is what
     /// emulates Hydra's behavior.
-    fn place_nix(&self) -> Result<(), std::io::Error> {
-        let mut file = File::create(self.outpath_nix_path())?;
-        file.write_all(include_bytes!("outpaths.nix"))?;
+    fn place_nix(&self) -> Result<(), Error> {
+        let outpath = self.outpath_nix_path();
 
-        Ok(())
+        let mut file = match File::create(&outpath) {
```
grahamc: how about `.map_err(|e| Error::CreateFile(outpath, e))?` or something to that effect? (and same feedback for other similar cases.)

cole-h: It has been done, my liege.

PR #457, opened 2020-04-07
More logging

cole-h:
Avoid cluttering logs further.

Also, split out the `info!` with a newline to make logging more consistent.

### DIFF
--- a/ofborg/src/tasks/build.rs
+++ b/ofborg/src/tasks/build.rs
@@ -373,7 +373,8 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
         };
 
         info!("ok built ({:?}), building", status);
-        info!("Lines:\n-----8<-----");
+        info!("Lines:");
+        info!("-----8<-----");
         actions
             .log_snippet()
             .iter()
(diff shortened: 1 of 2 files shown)


ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -568,6 +568,14 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
 fn request_reviews(maint: &maintainers::ImpactedMaintainers, pull: &hubcaps::pulls::PullRequest) {
     if maint.maintainers().len() < 10 {
         for maintainer in maint.maintainers() {
+            if let Ok(pr) = pull.get() {
```
grahamc: Let's move that `pull.get()` out of this loop :)

cole-h: Whoops, you're right.

ofborg/src/tasks/build.rs
```diff
@@ -373,7 +373,8 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
         };
 
         info!("ok built ({:?}), building", status);
-        info!("Lines:\n-----8<-----");
+        info!("Lines:");
+        info!("-----8<-----");
```
grahamc, line 377: Nice. Once we're sloggin', we could even pass the lines in as a context argument!

grahamc:
Thanks!

PR #458, opened 2020-04-09
{checkout,clone}: redirect output to /dev/null

cole-h:
It fills the logs with typically-useless information. We don't really
care about the novels people write in their commit messages.

The only `git` commands that were left unmodified are the ones where we
use their output in some way (and thus, it isn't printed to stdout/err
anyways).

---

There is a chance that this discards actual error information. However, in the past... *checks calendar* 4 days I've been a log-watcher, I've only seen innocuous output from `git` that might be useful or interesting if it was in an interactive session and not just a byproduct of our various checks.

### DIFF
--- a/ofborg/src/checkout.rs
+++ b/ofborg/src/checkout.rs
@@ -4,7 +4,7 @@ use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::io::{Error, ErrorKind};
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 pub struct CachedCloner {
     root: PathBuf,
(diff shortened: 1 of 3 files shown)


ofborg/src/checkout.rs
```diff
@@ -95,11 +95,14 @@ impl CachedProjectCo {
...
         let mut lock = self.lock()?;
 
+        info!("Fetching PR #{}", pr_id);
         let result = Command::new("git")
             .arg("fetch")
             .arg("origin")
             .arg(format!("+refs/pull/{}/head:pr", pr_id))
             .current_dir(self.clone_to())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
```
grahamc: I'm a _bit_ unsure about redirecting stderr too...I wonder how much of the output you're (rightly) trying to squelch here is in stdout vs stderr?

ofborg/src/clone.rs
```diff
@@ -112,20 +114,25 @@ pub trait GitClonable {
             .arg("am")
             .arg("--abort")
             .current_dir(self.clone_to())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
```
cole-h, line 118: Is redirecting both stdout and stderr here OK? stderr should always have something to say because we shouldn't still be in a `merge` or `am` session. Since I just want to  get rid of the stderr output, the stdout redirect is unnecessary.

EDIT: OK'd on IRC.

PR #459, opened 2020-04-12
Cachix

sorki:
I have no idea what I'm doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,39 @@ corruption.
 
 See also: https://github.com/NixOS/ofborg/wiki/Operating-a-Builder
 
+## Hacking
+
+```
+git clone https://github.com/NixOS/ofborg/
+cd ofborg
+nix-shell ./shell.nix
+cd ofborg # enter subdirectory with Rust code
+cargo build
+cargo check
+cargo test
+```
+
+Be sure to test your changes by running the following commands from the root of the repository:
+
+```
+nix-shell --run checkPhase -A mozilla-rust-overlay # checks rustfmt and Clippy
+nix-shell --run checkPhase # runs actual tests
+```
+
+Currently there's no way to set-up a testing instance easily.
+Send a PR if `check` and `test` are passing. Make sure
+to format your code with `cargo fmt` and check for additional
+warnings with `cargo clippy`.
+
+To disable warnings as errors, run your command with an empty `RUSTFLAGS`. For example:
+
+```
+RUSTFLAGS= cargo clippy
+```
+
+This is because `shell.nix` sets `RUSTFLAGS` to `-D warnings`,
+which tells clippy to treat warnings as errors.
+
 ## old php stuff...
 
 Only Graham needs to do this, since I run the only remaining PHP
--- a/ofborg/src/tasks/build.rs
+++ b/ofborg/src/tasks/build.rs
@@ -1,5 +1,6 @@
 use crate::checkout;
 use crate::commentparser;
+use crate::config::CachixConfig;
 use crate::message::buildresult::{BuildResult, BuildStatus, V1Tag};
 use crate::message::{buildjob, buildlogmsg};
 use crate::nix;
@@ -10,12 +11,14 @@ use amqp::protocol::basic::{BasicProperties, Deliver};
 use uuid::Uuid;
 
 use std::collections::VecDeque;
+use std::process::Command;
 
 pub struct BuildWorker {
     cloner: checkout::CachedCloner,
     nix: nix::Nix,
     system: String,
     identity: String,
+    cachix: Option<CachixConfig>,
 }
 
 impl BuildWorker {
@@ -24,12 +27,14 @@ impl BuildWorker {
         nix: nix::Nix,
         system: String,
         identity: String,
+        cachix: Option<CachixConfig>,
     ) -> BuildWorker {
         BuildWorker {
             cloner,
             nix,
             system,
             identity,
+            cachix,
         }
     }
 
@@ -112,6 +117,7 @@ impl<'a, 'b> JobActions<'a, 'b> {
             attempted_attrs: None,
             skipped_attrs: None,
             status: BuildStatus::Failure,
+            cachix_uploaded: None,
         };
 
         let result_exchange = self.result_exchange.clone();
@@ -193,6 +199,7 @@ impl<'a, 'b> JobActions<'a, 'b> {
             skipped_attrs: Some(not_attempted_attrs),
             attempted_attrs: None,
             status: BuildStatus::Skipped,
+            cachix_uploaded: None,
         };
 
         let result_exchange = self.result_exchange.clone();
@@ -219,6 +226,7 @@ impl<'a, 'b> JobActions<'a, 'b> {
         status: BuildStatus,
         attempted_attrs: Vec<String>,
         not_attempted_attrs: Vec<String>,
+        cachix_uploaded: bool,
     ) {
         let msg = BuildResult::V1 {
             tag: V1Tag::V1,
@@ -231,6 +239,7 @@ impl<'a, 'b> JobActions<'a, 'b> {
             status,
             attempted_attrs: Some(attempted_attrs),
             skipped_attrs: Some(not_attempted_attrs),
+            cachix_uploaded: Some(cachix_uploaded),
         };
 
         let result_exchange = self.result_exchange.clone();
@@ -346,6 +355,7 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
             return;
         }
 
+        // Nix build begins here
         let mut spawned =
             self.nix
                 .safely_build_attrs_async(refpath.as_ref(), buildfile, can_build.clone());
@@ -382,8 +392,53 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
             .last();
         info!("----->8-----");
 
-        actions.build_finished(status, can_build, cannot_build_attrs);
+        let cachix_result = match &self.cachix {
+            Some(conf) if status == BuildStatus::Success => cache_push(conf, actions.log_snippet()),
+            _ => {
+                info!("Skipping Cachix because it was either not configured, or the build failed");
+
+                false
+            }
+        };
+
+        actions.build_finished(status, can_build, cannot_build_attrs, cachix_result);
         info!("Done!");
+    }
+}
+
+fn cache_push(cfg: &CachixConfig, log: Vec<String>) -> bool {
+    if !log.is_empty() {
+        let store_path = log.last().unwrap();
+        let mut cmd = Command::new("cachix");
+
+        if let Some(path) = &cfg.config_path {
+            cmd.arg(format!("--config {}", path));
+        }
+
+        if let Some(host) = &cfg.host {
+            cmd.arg(format!("--host {}", host));
+        }
+
+        cmd.arg("push");
+
+        if let Some(jobs) = &cfg.push_jobs {
+            cmd.arg(format!("--jobs {}", jobs));
+        }
+
+        if let Some(level) = &cfg.push_compression_level {
+            cmd.arg(format!("--compression-level {}", level));
+        }
+
+        cmd.arg(&cfg.cache_name)
+            .arg(store_path)
+            .status()
+            .ok()
+            .map(|res| res.success())
+            .unwrap_or(false)
+    } else {
+        error!("Successful build with no output");
+
+        false
     }
 }
 
@@ -415,6 +470,7 @@ mod tests {
             nix,
             "x86_64-linux".to_owned(),
             "cargo-test-build".to_owned(),
+            None,
         );
 
         worker
--- a/ofborg/src/tasks/githubcommentposter.rs
+++ b/ofborg/src/tasks/githubcommentposter.rs
@@ -151,6 +151,14 @@ fn result_to_check(result: &LegacyBuildResult, timestamp: DateTime<Utc>) -> Chec
         ));
     }
 
+    if let Some(uploaded) = result.cachix_uploaded {
+        if uploaded {
+            summary.push(String::from("Build successfully uploaded to Cachix."));
+        } else {
+            summary.push(String::from("Failed to upload build to Cachix."));
+        }
+    }
+
     let text: String = if !result.output.is_empty() {
         let mut reply: Vec<String> = vec![];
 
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -229,6 +229,31 @@ corruption.
...
+
+```
+git clone https://github.com/NixOS/ofborg/
+cd ofborg
+nix-shell ./shell.nix
+cd ofborg
+cargo build
+cargo check
+cargo test
+```
```
cole-h, line 242: I'd also add running `nix-shell --run checkPhase --arg useNix1 false -A mozilla-rust-overlay` from the root of the repo (taken from `.travis.yml`) because that's where most of my build failures came from x)

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,59 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+
+            if let Some(level) = &cfg.push_compression_level {
+                cmd.arg(format!("--compression-level {}", level));
+            }
+
+            let result = cmd
+                .arg(cfg.cache_name)
+                .arg(store_path)
+                .status()
+                .expect("All done.");
```
cole-h: If an `expect` fires, it will result in a runtime panic. If this cannot be avoided, it would be nice to have a more descriptive message of what happened/went wrong. I might even be tempted to write the following for this branch:

```rust
            info!("Pushing to Cachix");
            cmd.arg(&cfg.cache_name)
                .arg(store_path)
                .status()
                .ok()
                .map(|res| res.success())
                .unwrap_or(false)
```

If the command failed, we return false. If the command succeeded, we return whether or not it was successful.

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,59 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+                false
+            }
+        };
+
+        actions.build_finished(status, can_build, cannot_build_attrs, cachix_result);
         info!("Done!");
     }
 }
 
+fn cache_build(cfg: CachixConfig, status: BuildStatus, log: Vec<String>) -> bool {
```
cole-h: Maybe `cache_push` would be a more appropriate name here. It doesn't look like we're building from the cache (though we are building *a* cache).

```suggestion
fn cache_push(cfg: &CachixConfig, status: BuildStatus, log: Vec<String>) -> bool {
```

My version of this function: https://gist.github.com/cole-h/00996da012d52ce8c95efb34743daf32

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,59 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
         info!("----->8-----");
 
-        actions.build_finished(status, can_build, cannot_build_attrs);
+        let cachix_result = match &self.cachix {
+            Some(conf) => cache_build(conf.clone(), status.clone(), actions.log_snippet()),
+            None => {
+                info!("Cachix not configured, skipping");
+                false
+            }
+        };
```
cole-h: Since `status` is only used to check if it was a success, there's no reason to make it an argument (and lets us avoid `clone`ing it) -- just check in the `match` branch itself.

```suggestion
        let cachix_result = match &self.cachix {
            Some(conf) if status == BuildStatus::Success => {
                cache_build(conf, actions.log_snippet())
            }
            _ => {
                info!("Skipping Cachix because it was either not configured, or the build failed");

                false
            }
        };
```

We can also avoid `clone`ing the config by making `cache_build` (or `cache_push`, if I have my way) take a reference to a `CachixConfig`.

cole-h (changes requested):
I know this is just a draft, but I have a few comments and suggestions. We'll also need to add `cachix` to the environment (maybe `shell.nix`) at some point.

ofborg/src/tasks/build.rs
sorki: Thanks, was struggling with this bit a bit.

ofborg/src/tasks/build.rs
sorki: It was meant as a verb like `cache that build` but I like `cache_push` better.

README.md
sorki: Done

sorki:
Thanks for the suggestions @cole-h. I've marked this ready for review now ;)

domenkozar:
cc @grahamc

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+cd ofborg # enter sub directory with Rust code
+cargo build
+cargo check
+cargo test
+```
+
+Make sure to also test with (from repository root):
+
+```
+nix-shell --run checkPhase --arg useNix1 false -A mozilla-rust-overlay
```
cole-h: Hehe, I forgot to come back and update my comment... I removed the Nix1 stuff in #460. It should just be necessary to run the following:

```suggestion
nix-shell --run checkPhase -A mozilla-rust-overlay
nix-shell --run checkPhase
```

The first one is our "Cargo pedantry" check (make sure Clippy and rustfmt are happy), while the second is what actually runs tests.

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+```
+
+
+Currently there's no way to set-up a testing instance easily.
+Send a PR if `check` and `test` are passing. Make sure
```
cole-h, line 252:
```suggestion
Currently, there's no easy way to set up a test instance. Feel free to
send a PR if `cargo check` and `cargo test` are passing. Make sure
```

(reflow where applicable)

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+to format your code with `cargo fmt` and check for additional
+warnings with `cargo clippy`.
+
+To disable warnings as errors run
+
+```
+unset RUSTFLAGS
+```
+
+which are set to `-D warnings` by `shell.nix`.
```
cole-h: I don't think this line is necessary... who cares what's set and where :P

ofborg/src/tasks/build.rs
```diff
@@ -346,6 +355,7 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
             return;
         }
 
+        // here we start the actual nix build
```
cole-h:
```suggestion
```

This comment feels unnecessary to me. The next ~20 lines give context that I feel make it obvious that we're starting the Nix build hereabouts (especially, `self.nix.safely_build_attrs_async` and `match s.code() { ... => BuildStatus::Success`).

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,55 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+        }
+
+        cmd.arg(&cfg.cache_name)
+            .arg(store_path)
+            .status()
+            .ok()
+            .map(|res| res.success())
+            .unwrap_or(false)
+    } else {
+        error!("Successful build with no output");
```
cole-h, line 439: Nit, but I think makes it more readable if we separate expressions from statements.

```suggestion
        error!("Successful build with no output");
        
```

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+
+Currently there's no way to set-up a testing instance easily.
+Send a PR if `check` and `test` are passing. Make sure
+to format your code with `cargo fmt` and check for additional
+warnings with `cargo clippy`.
+
+To disable warnings as errors run
+
+```
+unset RUSTFLAGS
```
cole-h: `unset` doesn't exist in the fish shell, so let's make this slightly more generalized:

```suggestion
To disable warnings as errors, run your command with an empty `RUSTFLAGS`. For example:

\```
RUSTFLAGS= cargo clippy
```

(I don't know how to escape triple backquotes inside triple backquotes, so I put a literal backslash in there, meaning you'll have to change this manually rather than just apply my suggestion...)

ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -151,6 +151,14 @@ fn result_to_check(result: &LegacyBuildResult, timestamp: DateTime<Utc>) -> Chec
         ));
     }
 
+    if let Some(up) = result.cachix_uploaded {
+        if up {
```
cole-h: More nits, but I'd like the longer and slightly more explicit variable name here... Feel free to disregard.

```suggestion
    if let Some(uploaded) = result.cachix_uploaded {
        if uploaded {
```

cole-h (changes requested):
A few more nits.

ofborg/src/tasks/build.rs
sorki: It makes it easier to spot / grep for nix build, it wasn't clear to me at first and there was at least one other person asking about this.

cole-h: Then how about:

```suggestion
        // Nix build begins here
```

?

README.md
sorki: Well I've had too look it up as it was trolling me :D

ofborg/src/tasks/build.rs
sorki: Sounds good!

README.md
```diff
@@ -229,6 +229,39 @@ corruption.
...
+cargo build
+cargo check
+cargo test
+```
+
+Make sure to also test with (from repository root):
+
+```
+nix-shell --run checkPhase -A mozilla-rust-overlay
+nix-shell --run checkPhase
```
cole-h: I figure it's better to explain what each command does in the README like I did earlier.

```suggestion
Be sure to test your changes by running the following commands from the root of the repository:

\```
nix-shell --run checkPhase -A mozilla-rust-overlay # checks rustfmt and Clippy
nix-shell --run checkPhase # runs actual tests
```

(again, reflow where appropriate and remove the backslash locally...)

README.md
```diff
@@ -229,6 +229,39 @@ corruption.
 
 See also: https://github.com/NixOS/ofborg/wiki/Operating-a-Builder
 
+## Hacking
+
+```
+git clone https://github.com/NixOS/ofborg/
+cd ofborg
+nix-shell ./shell.nix
+cd ofborg # enter sub directory with Rust code
```
cole-h:
```suggestion
cd ofborg # enter subdirectory with Rust code
```

README.md
```diff
@@ -229,6 +229,39 @@ corruption.
...
+cargo test
+```
+
+Make sure to also test with (from repository root):
+
+```
+nix-shell --run checkPhase -A mozilla-rust-overlay
+nix-shell --run checkPhase
+```
+
```
cole-h, line 250:
```suggestion
```

cole-h (changes requested):
A few more nits I missed the last time and LGTM.

README.md
cole-h: Well, if you want to leave it in, it'll need to be reworded, since it no longer makes sense with the current content ;)

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,56 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+
```
cole-h: There's a tab here from the previous suggestion in this area which will need to be fixed locally for `Cargo pedantry` to pass. Thanks GitHub.

EDIT: Specifically, there's a tab in the "blank" line between `error!` and `false`.

sorki: Fixed with `cargo fmt`

README.md
sorki: Now it is

By default `RUSTFLAGS` contain `-D warnings` set by `shell.nix`.

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+nix-shell ./shell.nix
+cd ofborg # enter subdirectory with Rust code
+cargo build
+cargo check
+cargo test
+```
+
+Be sure to test your changes by running the following commands from the root of the repository:
+
+\```
```
cole-h: Don't forget these backslashes (they were just there to let the suggestions block be properly actionable).

README.md
```diff
@@ -229,6 +229,38 @@ corruption.
...
+```
+
+Currently, there's no easy way to set up a test instance. Feel free to
+send a PR if `cargo check` and `cargo test` are passing. Make sure
+to format your code with `cargo fmt` and check for additional
+warnings with `cargo clippy`.
+
+To disable warnings as errors, run your command with an empty `RUSTFLAGS`. For example:
+
+\```
```
cole-h: Don't forget these backslashes (they were just there to let the suggestions block be properly actionable).

README.md
sorki: I see, was wondering about it :)

README.md
cole-h: "This is because `shell.nix` sets `RUSTFLAGS` to `-D warnings`, which tells clippy to treat warnings as errors." sounds better, IMHO. Otherwise, LGTM.

EDIT: added "This is because" to the beginning -- either way works (with or without those 3 words).

sorki: Thanks!

cole-h (approved):
LGTM! Thanks for dealing with my nitpicks :)

ofborg/src/tasks/build.rs
```diff
@@ -382,11 +392,56 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
...
+        };
+
+        actions.build_finished(status, can_build, cannot_build_attrs, cachix_result);
         info!("Done!");
     }
 }
 
+fn cache_push(cfg: &CachixConfig, log: Vec<String>) -> bool {
+    if !log.is_empty() {
+        let store_path = log.last().unwrap();
```
cole-h, line 411: Something I just thought of that should probably be addressed: what about when there were multiple builds, e.g. here: https://logs.nix.ci/?key=nixos/nixpkgs.82982&attempt_id=62265253-63b3-44e2-aa81-3fb8464be6ff

```
/nix/store/vb15nidh4wbsg9lrf93iv859mvkbn0pl-ripgrep-12.0.1
/nix/store/vywixxgrbrq76nbdra96h3nzrlnagq5m-zcash-2.1.1-1
/nix/store/v9c8cyl9rrspkv18w0f6b56ydb0350gf-lsd-0.17.0
```

In this case, I believe only the `lsd` path will be pushed to the cache.

cole-h (changes requested):

ofborg/src/tasks/build.rs
sorki: Good point! We can either call `cache_push` for each path or pipe them to Cachix via `stdin`. Just need to identify all the paths - maybe something like get all the last few consecutive lines matching `/nix/store` prefix?

cole-h: Or, get the last `n` lines, where `n = number of built attrs`? We gate this on a successful build anyways, so this should be a safe bet (assuming that any failing build sets status to be not-`BuildStatus::Success` -- I'll have to read through the source again to be entirely certain).

Or both, as sanity checks for the other?

I'll do some brainstorming.

sorki:
I decided I don't care.

domenkozar:
It would be great if binaries were pushed and available to users for testing.

I'm happy to sponsor a cache for ofborg if someone manages to get integration working.

PR #466, opened 2020-04-23
config.public.json: add @ThomasMader

ThomasMader:
I would like to build on Darwin again.
I work on D programming language (dlang) stuff from time to time and as of lately I try to support others who are about to contribute something for dlang.

For example I am maintainer for ldc and dmd.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -95,6 +95,7 @@
             "srhb",
             "talyz",
             "teto",
+            "ThomasMader",
             "thoughtpolice",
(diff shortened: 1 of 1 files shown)


grahamc:

veprbl:
This one was targeting trusted_users

ThomasMader:
Exactly, I wanted to become a trusted user so I can build on Darwin again.
Should I open a new PR?

PR #471, opened 2020-04-25
carnix workspace

LnL7:


ofborg/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "ofborg"
 version = "0.1.8"
 authors = ["Graham Christensen <graham@grahamc.com>"]
-include = ["Cargo.toml", "Cargo.lock", "src", "test-srcs", "build.rs"]
+include = ["Cargo.toml", "ofborg"]  # TODO remove when carnix is fixed
```
cole-h, line 5: What needs to be "fixed" in Carnix, and what would be removed? The entire `include` line? Just the `ofborg` folder from the includes?

LnL7: Take a look at https://github.com/NixOS/ofborg/pull/471/commits/96639e9d7e456bcee261d40b7d92c6043ea9f907, that only contains the workaround. There are 2 issues I noticed.

1. the includes for a member have a space which means the filter doesn't match anything and nix filters the sources out
2. specified includes are not relative to the member and declaring nested paths doesn't work either.

PR #474, opened 2020-04-28
lapin amqp implementation

LnL7:
An attempt at switching the amqp library to improve hearbeats.


ofborg/src/easylapin.rs
```diff
@@ -0,0 +1,173 @@
...
+use async_std::stream::StreamExt;
+use async_std::task;
+use lapin::{
+    message::Delivery, options::*, types::FieldTable, BasicProperties, Channel, CloseOnDrop,
+    Connection, ConnectionProperties, ExchangeKind,
+};
+
+pub fn from_config(cfg: &RabbitMQConfig) -> Result<CloseOnDrop<Connection>, lapin::Error> {
+    let opts = ConnectionProperties::default();
+    // TODO version
```
grahamc: We'll definitely want this to-done for merge

grahamc (approved):
Those TODO's should  definitely be done first. The only other thing I noticed is the unwrap's, but the current API design really does not tolerate error handling well and so I'm going to ignore that for now.

ofborg/src/bin/builder.rs
```diff
@@ -101,9 +100,8 @@ fn main() {
         .unwrap();
 
     info!("Fetching jobs from {}", &queue_name);
-    channel.start_consuming();
-    channel.close(200, "Bye").unwrap();
-    info!("Closed the channel");
-    session.close(200, "Good Bye");
+    task::block_on(handle);
+
+    drop(conn);
```
cole-h: Is there a reason this needs to be explicitly dropped, rather than letting it just fall out of scope?

Asked on IRC, but just wanted to make sure it got answered -- maybe it's a quirk of async, or something.

LnL7: Nope, the type is `CloseOnDrop<Connection>` 😄, but not dropping explicitly would mean the log message comes too early.

cole-h: A comment to that effect would be nice, so nobody accidentally removes it :eyes:

ofborg/src/easylapin.rs
```diff
@@ -0,0 +1,198 @@
+use std::pin::Pin;
+
+use crate::config::RabbitMQConfig;
+use crate::easyamqp::*;
```
cole-h: Any chance this could be un-globbed? A nit, but I like to be able to see where everything is coming from :^)

cole-h (approved):

grahamc:
:partying_face:

PR #476, opened 2020-04-29
config.public.json: add cole-h to trusted_users

cole-h:
This will let me build packages like `fish` on darwin, to try and catch
any new regressions (there's been one in each of the past two bumps I've
submitted...).


cole-h:
Thanks!

PR #479, opened 2020-05-01
remove static lifetimes from easylapin

LnL7:
Unlike amqp::Consumer lapin channels don't need to have a 'static
lifetime, the workers just need to live at least as long as the channel.


cole-h (approved):
LGTM.

PR #481, opened 2020-05-05
skip stdenv rebuilds

LnL7:
Before attempting builds check if the stdenv for this branch is
available in the cache.  If this is not the case then either the merge
request is a mass rebuild or it's targeting a branch like staging.

### DIFF
--- a/ofborg/src/tasks/build.rs
+++ b/ofborg/src/tasks/build.rs
@@ -325,10 +325,7 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
             return;
         }
 
-        info!(
-            "Got path: {:?}, determining which ones we can build ",
-            refpath
-        );
+        info!("Determining which attributes we can build");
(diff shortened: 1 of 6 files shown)


ofborg/src/tasks/build.rs
```diff
@@ -341,6 +338,22 @@ impl notifyworker::SimpleNotifyWorker for BuildWorker {
             .map(|(attr, _)| attr)
             .collect();
 
+        info!("Checking for stdenv rebuild");
+        let in_cache = self
+            .nix
+            .safely_query_cache_for_attr(refpath.as_ref(), buildfile, "stdenv".to_string())
```
cole-h: Nit, but I prefer `String::from()` over `to_string()` for static strings like this.

(Though AFAICT they are the same; personal preference, really.)

```suggestion
            .safely_query_cache_for_attr(refpath.as_ref(), buildfile, String::from("stdenv"))
```

PR #482, opened 2020-05-06
"Not you: team." (Rework README)

### LINKED ISSUE #445: Fix trusted/known users description in README
@bhipple and @jonringer tell me that linux builds happen by default now, so the README needs to be updated.

cole-h:
Some of the information in the README is outdated, such as the sections
detailing the privileges of known users (everybody is now a known user).
The only distinction made is between that of trusted users and everybody
else. Trusted users gain an additional host to run builds and tests on:
x86_64-darwin. Everybody else is locked to x86_64-linux and
aarch64-linux.

With the removal of all mentions of "known users", the
`config.{extra-,}known-users.json` files have also been removed. This
change will require coordination with the infrastructure repo [1] --
namely, a PR removing references to these files (otherwise, I imagine,
deploys will fail).

Something Graham has mentioned a few times is how he wants to decouple
ofborg from his persona. Thus, `@ofborg` is now used in place of
`@GrahamcOfBorg` everywhere possible (it is still possible to use
`@GrahamcOfBorg`, so it has not been removed from the codebase at this
time).

[1] https://github.com/ofborg/infrastructure

---

Thanks to @sorki, there is now a `Hacking` section that details a
quick-and-dirty way to get started with working on ofborg's internals.

Depends on https://github.com/ofborg/infrastructure/pull/21.

Closes #445.

EDIT: I've removed the "Running a Build" section from the README and placed its contents in the Wiki: https://github.com/NixOS/ofborg/wiki/Operating-a-Builder/


grahamc (approved):
Yes! Thank you! (cc @andir for title.)

PR #487, opened 2020-05-14
lapin github-comment-poster

LnL7:


ofborg/src/tasks/githubcommentposter.rs
```diff
@@ -50,22 +50,33 @@ impl worker::SimpleWorker for GitHubCommentPoster {
...
                 for architecture in queued_job.architectures.iter() {
                     checks.push(job_to_check(&queued_job.job, &architecture, Utc::now()));
                 }
+                queued_job.job.pr.to_owned()
             }
             PostableEvent::BuildFinished(finished_job) => {
                 let result = finished_job.legacy();
                 repo = result.repo.clone();
                 checks.push(result_to_check(&result, Utc::now()));
+                finished_job.pr()
```
cole-h, line 65: Does this need to be a function, considering you have `queued_job.job.pr.to_owned()` in the other branch? For something one-shot like this appears to be, why not just `finished_job.pr.to_owned()` or something?

LnL7: It's an enum, otherwise you'd have to match on all the cases here to extract it.

```
E0609: no field `pr` on type `&message::buildresult::BuildResult`
```

cole-h: Brain is slow. In my defense, I just came out of a final...

Ack.

cole-h (approved):

PR #488, opened 2020-05-15
don't request maintainer reviews if many files changed

LnL7:
This either indicates it's most likely either a tree-wide change,
staging-next merge or just a pull request with an incorrect target
branch.  For none of these cases are usually relevant for specific
package maintainers.

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -24,8 +24,10 @@ use hubcaps::checks::{CheckRunOptions, CheckRunState, Conclusion, Output};
 use hubcaps::gists::Gists;
 use hubcaps::issues::{Issue, IssueRef};
 use hubcaps::repositories::Repository;
-use tracing::warn;
+use tracing::{info, warn};
 use uuid::Uuid;
+
+static MAINTAINER_REVIEW_MAX_CHANGED_PATHS: usize = 64;
 
 pub struct NixpkgsStrategy<'a> {
     job: &'a EvaluationJob,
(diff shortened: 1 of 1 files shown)


ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -270,6 +270,22 @@ impl<'a> NixpkgsStrategy<'a> {
                 },
             );
 
+            if changed_paths.len() > 64 {
```
cole-h: Would this be a good thing to make `const`? e.g. `const MAX_CHANGED_PATHS: u32 = 64;`

Additionally, is there a reason 64 was chosen? Or is it just an arbitrary number?

ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -270,6 +270,22 @@ impl<'a> NixpkgsStrategy<'a> {
...
+                info!(
+                    "pull request has {} changed files, skipping review requests",
+                    changed_paths.len()
+                );
+                let status = CommitStatus::new(
+                    self.repo.statuses(),
+                    self.job.pr.head_sha.clone(),
+                    String::from("grahamcofborg-eval-check-maintainers"),
+                    String::from("large change, skipping maintainer review"),
```
cole-h: I like the description from the `info!` above.

```suggestion
                    String::from("large change, skipping review requests"),
```

Maybe also `s/large/massive/`? ("mass rebuild", "massive change")

ofborg/src/tasks/eval/nixpkgs.rs
LnL7: Mostly arbitrary, half is already seems rather large so I'm thinking this probably won't skip any cases that would be valid.

ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -270,6 +272,22 @@ impl<'a> NixpkgsStrategy<'a> {
                 },
             );
 
+            if changed_paths.len() > MAINTAINER_REVIEW_MAX_CHANGED_PATHS {
+                info!(
+                    "pull request has {} changed files, skipping review requests",
```
cole-h: Nit; might as well stay consistent with the variable's name.

```suggestion
                    "pull request has {} changed paths, skipping review requests",
```

cole-h (approved):

PR #489, opened 2020-05-20
packet exporter

LnL7:


packet-exporter/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+extern crate prometheus;
+
+use std::error::Error;
+
+use hyper::header::CONTENT_TYPE;
+use hyper::service::{make_service_fn, service_fn};
+use hyper::{Body, Client, Request, Response, Server, Uri};
+use hyper_tls::HttpsConnector;
+use prometheus::{Encoder, GaugeVec, TextEncoder};
+use serde_derive::{Deserialize, Serialize};
```
cole-h: Since we're already using 2018 edition, I'd like to keep the occurrences of `#[macro_use]` as low as possible.

```suggestion
use std::error::Error;

use hyper::header::CONTENT_TYPE;
use hyper::service::{make_service_fn, service_fn};
use hyper::{Body, Client, Request, Response, Server, Uri};
use hyper_tls::HttpsConnector;
use lazy_static::lazy_static;
use prometheus::{register_gauge_vec, Encoder, GaugeVec, TextEncoder};
use serde_derive::{Deserialize, Serialize};
```

cole-h (changes requested):

packet-exporter/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let response = Response::builder()
+        .status(200)
+        .header(CONTENT_TYPE, encoder.format_type())
+        .body(Body::from(buffer))?;
+
+    Ok(response)
+}
+
+#[tokio::main]
```
LnL7: FYI I also looked into tide/surf which don't need tokio, but ran into some issues with carnix there. Have some more patches ready if we want change the metrics of the other ofborg services.

cole-h: What kind of changes to the metrics did you have in mind?

LnL7: With prometheus you ideally want to pull the current state of metrics directly from services instead of pushing data as events, which is what the stats currently do.
That's not always possible but if it's much more flexible since you don't need to deal with the caveats for push based data anymore.

cole-h: Oh right. I think we talked about this in IRC yesterday/some time recently. SGTM :+1: 

Re tide/surf: I think tokio is fine (although I have very little experience in the async and HTTP client areas). Why is the lack of tokio desirable? Smaller dependency graph/quicker compile times?

LnL7: Yeah, tokio had a noticable build time impact last time I checked. Not a huge deal, but we should only use one runtime per crate IMHO so I would want to replace async-std in that case.

cole-h: Makes sense. Are the issues with Carnix something that could be patched upstream (like we did for the workspace stuff)? 

Maybe just leave a FIXME or something if you don't get to it in this PR, and it can be revisited later on down the line.

LnL7: Yeah, [carnix:48](https://nest.pijul.com/pmeunier/carnix/discussions/48) is one of them :smile:, but I got stuck in some kind of features mess.

LnL7:
Moved to https://github.com/LnL7/packet-exporter.

PR #493, opened 2020-05-21
lapin stats

LnL7:


ofborg/src/bin/stats.rs
```diff
@@ -1,96 +1,83 @@
...
-        Server::http(addr)
-            .unwrap()
-            .handle(move |_: Request, res: Response| {
-                res.send(metrics.prometheus_output().as_bytes()).unwrap();
-            })
-            .unwrap();
+        Server::http(addr)?.handle(move |_: Request, res: Response| {
+            res.send(metrics.prometheus_output().as_bytes()).unwrap();
+        })?;
+        Ok::<_, Box<dyn Error + Sync + Send + '_>>(())
```
cole-h, line 74: This annotation is pretty gnarly. I also saw this in the Packet exporter (https://github.com/NixOS/ofborg/pull/489/files#diff-02fbe59114407226c4d138e271cd71a5R76) -- maybe this could be a type alias? I don't know if something like `type Error = Box<dyn std::error::Error + Sync + Send + '_>;` is possible, but it looks nicer.

However, not a blocker. Do what you see fit.

cole-h (approved):

ofborg/src/bin/stats.rs
LnL7: I don't think you can for the lifetime, since that's context based. But it `BoxError<'_>` should work with `type BoxError<'a> = Box<dyn Error + Sync Send + 'a>;`

PR #495, opened 2020-05-22
convert log-message-collector to easylapin

LnL7:


ofborg/src/bin/log-message-collector.rs
```diff
@@ -1,78 +1,72 @@
...
-    channel
-        .bind_queue(easyamqp::BindQueueConfig {
+    chan.bind_queue(easyamqp::BindQueueConfig {
+        queue: queue_name.clone(),
+        exchange: "logs".to_owned(),
+        routing_key: Some("*.*".to_owned()),
+        no_wait: false,
+    })?;
+
+    let handle = chan.consume(
```
LnL7, line 52: unlike the other services this did prefetch, given the throughput of logs this probably shouldn't handle them one by one.

ofborg/src/easylapin.rs
```diff
@@ -117,6 +115,20 @@ impl<'a, W: SimpleWorker + 'a> ConsumerExt<'a, W> for CloseOnDrop<Channel> {
     }
 }
 
+// Same as a regular channel, but without prefetching,
+// used for services with multiple instances.
```
cole-h: Might as well make this a doc comment.

```suggestion
/// Same as a regular channel, but without prefetching,
/// used for services with multiple instances.
```

cole-h (approved):

PR #497, opened 2020-05-23
relax PullRequestAction parsing

LnL7:
Unknown events are fine here, as long as it's a valid event continue and
consider it not interesting instead.

Should alleviate fixes like #496 in the future.


cole-h (approved):
What's the reason for dropping the other variants? Is it because they are currently unused?

LnL7:
Yeah they where there for the same reason as ConvertToDraft, if we ever want to use specific actions they can be added back. For example with skipping builds for draft pull requests making the pr as ready should trigger builds.

PR #498, opened 2020-05-23
lapin: 1.0.0-beta4 -> 1.0.0-rc6

LnL7:

PR #499, opened 2020-05-24
use crate2nix

LnL7:
We've run into various issues with carnix, some have been patched already but I'm not sure how to fix other issues like #498. Crate2nix seems to handle all of the cases we have ran into so far.


andir:
Maybe use `cargo update` or `cargo lock` instead of the `cargo build`
invocation in nix/update-crates.sh as otherwise you impurely require all
the build inputs in the shell.

LnL7:
Yeah, but this is how it was before. Doesn't it make sense to ensure everything actually builds before updating?

andir:
I would argue it is time to fix that nix-shell shebang to contain the build
dependencies.

cole-h (changes requested):
Probably want to remove Carnix from `nix/overlay.nix` and `shell.nix`, as well:

https://github.com/NixOS/ofborg/blob/65db4d0da04f725ccbb2d92efa3fb69f0c5a8625/nix/overlay.nix#L3

https://github.com/NixOS/ofborg/blob/65db4d0da04f725ccbb2d92efa3fb69f0c5a8625/shell.nix#L86

cole-h (approved):
> +5,293 −7,724 

Nice! LGTM. Hopefully this requires less patching than Carnix did ;)

PR #500, opened 2020-05-24
lapin test binaries

LnL7:
Last bit of the transition.

### DIFF
--- a/ofborg/src/bin/build-faker.rs
+++ b/ofborg/src/bin/build-faker.rs
@@ -1,24 +1,25 @@
+use std::env;
+use std::error::Error;
+
+use async_std::task;
+use lapin::message::Delivery;
+use lapin::BasicProperties;
+
 use ofborg::commentparser;
 use ofborg::config;
-use ofborg::easyamqp;
+use ofborg::easylapin;
 use ofborg::message::{buildjob, Pr, Repo};
-use ofborg::notifyworker::{self, NotificationReceiver};
+use ofborg::notifyworker::NotificationReceiver;
 use ofborg::worker;
 
-use std::env;
-
-use tracing::info;
-
-fn main() {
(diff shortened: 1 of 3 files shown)


ofborg/src/bin/build-faker.rs
```diff
@@ -57,8 +66,5 @@ fn main() {
         }
     }
 
-    channel.close(200, "Bye").unwrap();
-    info!("Closed the channel");
-    session.close(200, "Good Bye");
-    info!("Closed the session... EOF");
+    Ok(())
```
cole-h, line 69: Don't need to `drop(conn);` here?

ofborg/src/bin/log-message-generator.rs
```diff
@@ -1,26 +1,38 @@
...
-    let mut session = easyamqp::session_from_config(&cfg.rabbitmq).unwrap();
-    info!("Connected to rabbitmq");
+    let arg = env::args()
+        .nth(1)
+        .expect("usage: log-message-generator <config>");
+    let cfg = config::load(arg.as_ref());
 
-    info!("About to open channel #1");
-    let mut chan = session.open_channel(1).unwrap();
+    let conn = easylapin::from_config(&cfg.rabbitmq)?;
```
cole-h, line 24: Same, don't need to drop this `conn`?

cole-h (approved):
Couple questions (well one question, occurring twice), but overall LGTM.

ofborg/src/bin/log-message-generator.rs
LnL7: The drop is there in the other binaries for logging, since this just runs and exits that doesn't seem particularly relevant here.

cole-h: Makes sense, thanks for the explanation.

PR #502, opened 2020-05-24
remove openssl 1.0.2

LnL7:


ofborg/Cargo.toml
```diff
@@ -21,7 +19,7 @@ tracing-subscriber = { version = "0.2.5", features = ["json"] }
 hubcaps = { git = "https://github.com/grahamc/hubcaps.git" }
 #hubcaps = { path = "../../../grahamc/hubcaps/" } # for testing patches
 hyper = "0.10.*"
-hyper-native-tls = "0.2.4"
+hyper-native-tls = "0.3.0"
```
cole-h, line 22: For posterity's sake: CI was failing because hyper-native-tls 0.2.4 depends on native-tls 0.1.5, which depends on openssl 0.9.24, which only supports OpenSSL up to 1.1.0 (NixOS is on 1.1.1g at the time of writing).

As you can see here, the solution was to bump hyper-native-tls to 0.3.0, which depends on native-tls 0.2.4, which depends on openssl 0.10.29, which *does* support OpenSSL 1.1.1g.

cole-h (approved):

cole-h:
🎉🎉

PR #507, opened 2020-05-25
async-std: 1.6.0 -> 1.5.0

LnL7:
Seems to cause an issue with the executors somehow.

    thread 'main' panicked at 'cannot run an executor inside another executor', <::std::macros::panic macros>:2:4


cole-h:
Potentially related: https://github.com/async-rs/async-std/issues/795.

cole-h (approved):

PR #511, opened 2020-05-28
move lib-tests to a build

LnL7:
This still has problems with rebuilds since it uses the target nixpkgs
to build the tests, but it's a regular build job which doesn't block the
entire evaluator anymore. Mitigates the main problem until we decide
what to do for these kind of use cases.

### DIFF
--- a/ofborg/src/commentparser.rs
+++ b/ofborg/src/commentparser.rs
@@ -72,6 +72,7 @@ pub enum Instruction {
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub enum Subset {
+    LibTests,
     Nixpkgs,
     NixOS,
 }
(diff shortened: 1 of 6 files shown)


infinisil:
See https://github.com/NixOS/ofborg/pull/472 which I think is a good solution in this specific case, since the lib tests only need some very basic builders from pkgs

cole-h (approved):
> until we decide what to do for these kind of use cases.

Is there a tracking issue filed for this? If not, could I impose on you to do so, at some point? :-)

LnL7:
Not sure there's anything specific, but #513 is probably most relevant.

wolfgangwalther:
lib-tests are now built in nixpkgs' CI.

Issue #148, opened 2018-03-26
Weird documentation tagging


srhb:
(Very low priority, mostly an observation)

In this PR, gcofborg decided that it needed a documentation tag. It's not obvious why this was the case.

https://github.com/NixOS/nixpkgs/pull/37879

Mic92:
Is the `tarsum` path used somewhere in the documentation?

srhb:
Possibly, but not in an obvious way that I can find. That would explain things, though.

Mic92:
Ok. tags should be based on the path. So I do not have an idea why it is added here: https://github.com/NixOS/ofborg/blob/0065a134617b1e7ce0d93479114ba102ed864b8c/config.public.json#L150

pbogdan:
Seems as if it's searching for `doc` in the whole string / path (not just at the beginning) and finds it in `docker`.

FRidh:
@grahamc how is the matching done? It should do the matching from the start of the string.

Folders should end with a `/`.

wolfgangwalther:
Those labels are now managed by actions/labeler in nixpkgs.

PR #515, opened 2020-06-15
config.public.json: add trailing slash to doc

zowoq:
Might help with #148?

I went through the last few pages of PRs with this label and it looks evenly split between PRs that have docs and not.

### DIFF
--- a/ofborg/src/tagger.rs
+++ b/ofborg/src/tagger.rs
@@ -255,7 +255,7 @@ impl PathsTagger {
             .possible
             .iter()
             .filter(|&(ref tag, ref _paths)| !self.selected.contains(&tag))
-            .filter(|&(ref _tag, ref paths)| paths.iter().any(|tp| path.contains(tp)))
+            .filter(|&(ref _tag, ref paths)| paths.iter().any(|tp| path.starts_with(tp)))
             .map(|(tag, _paths)| tag.clone())
             .collect();
         self.selected.append(&mut tags_to_add);
--- a/config.public.json
+++ b/config.public.json
@@ -239,7 +239,7 @@
             "doc/manual/release-notes/"
         ],
         "8.has: documentation": [
-            "doc",
+            "doc/",
             "nixos/doc"
         ],
         "8.has: module (update)": [


cole-h (approved):
SGTM for now, but I believe this will still trigger on bumps for packages e.g. `scdoc` (`pkgs/tools/typesetting/sc**doc/**default.nix`). It'll definitely cut down on false-positives, but at some point we might want to rework the parsing of these paths -- wherever that occurs -- to incorporate FRidh's suggestion from the linked issue (namely: "It should do the matching from the start of the [path]").

LnL7:
Perhaps we should also change this to absolute paths relative to the git checkout.

zowoq:
Changed to `starts_with` as suggested on `#nixos-ofborg`.

Seems matrix won't let me log in today so I can't comment there or `nixos-dev`.

ofborg/src/tagger.rs
```diff
@@ -255,7 +255,7 @@ impl PathsTagger {
             .possible
             .iter()
             .filter(|&(ref tag, ref _paths)| !self.selected.contains(&tag))
-            .filter(|&(ref _tag, ref paths)| paths.iter().any(|tp| path.contains(tp)))
+            .filter(|&(ref _tag, ref paths)| paths.iter().any(|tp| path.starts_with(tp)))
```
cole-h, line 258: I don't actually know if the path is only something like `pkgs/applications/...` or if it also includes the checkout (e.g. `~/workspace/vcs/nixpkgs` in my case).

@LnL7 since you run the darwin builder, do you know/can you verify if this is or isn't the case? (I still haven't taken the time to setup my own builder to play around with stuff like this... I'll do that Soon™.)

If it does start with the checkout path, we'll want to strip that out/not care about it before `starts_with` will work as expected.

zowoq:
Asking an unrelated question here as matrix still won't let me in. 

Now that `@GrahamcOfBorg` is deprecated, is there a reason for using `@ofborg` instead of switching to another prefix, e.g. `/ofborg` or `!ofborg` ?

The way `@` completes with users or teams that you don't actually want is a bit fiddly, it would be nice to change to something else if we can.

PR #522, opened 2020-06-19
config.public.json: add danieldk to trusted users

danieldk:
I am not sure when one is trusted enough to be a trusted user ;), but it would be nice to be able to test Darwin when reviewing PRs.


cole-h:

Cole

PR #526, opened 2020-07-24
config.public.json: Rename the GNOME label

jtojnar:
GNOME might switch to different versioning scheme soon so gnome3 might not make sense.

cc @cole-h @worldofpeace

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -137,7 +137,7 @@
         "6.topic: fetch": [
             "pkgs/build-support/fetch"
         ],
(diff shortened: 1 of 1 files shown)


jtojnar:
We probably need to synchronize the ofborg deploy with the renaming of the label in the nixpkgs repo to avoid splitting the tag.

cole-h (approved):
https://logs.nix.samueldr.com/nixos-dev/2020-07-24#3774790 (related IRC conversation, if you can call it that)

PR #527, opened 2020-08-05
ofborg: post a finished check if evaluation starts

Mic92:
We need this for better integration into github.

### DIFF
--- a/ofborg/src/tasks/evaluate.rs
+++ b/ofborg/src/tasks/evaluate.rs
@@ -32,6 +32,23 @@ pub struct EvaluationWorker<E> {
     tag_paths: HashMap<String, Vec<String>>,
 }
 
+/// Creates a finished github check, indicating that the ofborg eval started.
+/// This is a workaround github REST api, that only exposes finished external checks.
+/// We use the check status in our "Waiting for github action"
+fn post_eval_started_status(
+    api: hubcaps::statuses::Statuses,
+    commit: String,
+) -> Result<(), EvalWorkerError> {
+    Ok(CommitStatus::new(
+        api,
+        commit,
+        "ofborg-eval-started".to_owned(),
+        "Evaluation started".to_owned(),
+        None,
+    )
+    .set(hubcaps::statuses::State::Success)?)
+}
+
 impl<E: stats::SysEvents> EvaluationWorker<E> {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
@@ -314,6 +331,8 @@ impl<'a, E: stats::SysEvents + 'static> OneEval<'a, E> {
         );
 
         overall_status.set_with_description("Starting", hubcaps::statuses::State::Pending)?;
+
+        post_eval_started_status(repo.statuses(), job.pr.head_sha.clone())?;
 
         evaluation_strategy.pre_clone()?;
 


Mic92:
I assume I need to fix some tests since this adds a new check.

cole-h:
This wouldn't give a PR the "all clear" (green check) because `overall_status` would still be pending, correct?

Mic92:
> This wouldn't give a PR the "all clear" (green check) because `overall_status` would still be pending, correct?

Exactly it's posted right after the initial pending status.

cole-h (approved):
Should be fine, but I'd like to hear what @grahamc has to say (if anything).

zowoq:
Didn't Graham say he'd be AFK for a while this month? Hasn't been active on github or irc for a week.

Mic92:
cc @LnL7

zowoq:
https://github.com/NixOS/nixpkgs/commit/5ca7be38885382f0ea6d865544674c7b8ef705a2

I've bumped the timeout from 30min to 45min as we're still getting builds going red.

Update: I've just reset a bunch of PRs that timed out at 45mins.

cole-h:
Just a quick note: I'm in favor of merging this, but I just tried to redeploy to get our 3rd evaluator back, which failed.

Even if I did merge this, nothing would change until Graham has time to look into the deployment problem, so I'll just wait for him to be available again and 1) have him look at that; and 2) comment on this PR (if necessary).

zowoq:
https://github.com/NixOS/nixpkgs/pull/95126

Lets disable the actions until this is resolved so we don't have pages of red PRs.

zowoq:
What do we think about using actions to manually set a pending status without a timeout that is then cleared as part of this step?

There would be two actions statuses, one for the pending job and another from the job that sets it.

Mic92:
> What do we think about using actions to manually set a pending status without a timeout that is then cleared as part of this step?
> 
> There would be two actions statuses, one for the pending job and another from the job that sets it.

That would work too and is probably the better solution.

Mic92:
> > What do we think about using actions to manually set a pending status without a timeout that is then cleared as part of this step?
> > There would be two actions statuses, one for the pending job and another from the job that sets it.
> 
> That would work too and is probably the better solution.

I am not quite sure if it is possible so because the github token in github actions comes from the user that opens the pull request.

zowoq:
> https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
>
> This event is similar to `pull_request`, except that it runs in the context of the base repository of the pull request, rather than in the merge commit. This means that you can more safely make your secrets available to the workflows triggered by the pull request, because only workflows defined in the commit on the base repository are run.

Hopefully this would allow us to use a token in PRs?

Mic92:
> > [docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target)
> > This event is similar to `pull_request`, except that it runs in the context of the base repository of the pull request, rather than in the merge commit. This means that you can more safely make your secrets available to the workflows triggered by the pull request, because only workflows defined in the commit on the base repository are run.
> 
> Hopefully this would allow us to use a token in PRs?

Yes. This sounds great. We could also use this for cachix. We could even implement this without using ofborg:

1. Having a workflow that just creates a `Waiting for ofborg task` 
2. Use a second action that mark the first as complete if the ofborg suite is complete https://docs.github.com/en/actions/reference/events-that-trigger-workflows#check_suite

Mic92:
I will update this PR to mark the `Wait for ofborg` as non-pending once this project is maintained again.

worldofpeace:
ping @grahamc

zowoq:
@grahamc If we're going to start running more actions in PRs it would be nice to get this resolved.

zowoq:
Ping @cole-h @grahamc 

cc @SuperSandro2000

SuperSandro2000:
Why did you ping me?

zowoq:
Sorry, meant to link to IRC. This PR would address your comment about `wait for ofborg`: https://logs.nix.samueldr.com/nixos-borg/2021-01-27#4546552

ofborg/src/tasks/evaluate.rs
```diff
@@ -32,6 +32,23 @@ pub struct EvaluationWorker<E> {
     tag_paths: HashMap<String, Vec<String>>,
 }
 
+/// Creates a finished github check, indicating that the ofborg eval started.
+/// This is a workaround github REST api, that only exposes finished external checks.
+/// We use this check to in our "Waiting for github action"
```
andir: What is this line supposed to mean? It seems like the comment wasn't really finished?

Mic92: fixed.

andir (approved):
It is a bit unfortunate that we need this bit of code at all. Thanks to GitHubs limited external CI API. That being said this looks simple enough and solves a real-world pain that several people have noticed and disliked (myself included).

Until we have a way to signal that we started doing our job this seems like the best solution.

worldofpeace (approved):
If yall don't merge this :rofl:

ofborg/src/tasks/evaluate.rs
```diff
@@ -32,6 +32,23 @@ pub struct EvaluationWorker<E> {
...
+/// This is a workaround github REST api, that only exposes finished external checks.
+/// We use the check status in our "Waiting for github action"
+fn post_eval_started_status(
+    api: hubcaps::statuses::Statuses,
+    commit: String,
+) -> Result<(), EvalWorkerError> {
+    Ok(CommitStatus::new(
+        api,
+        commit,
+        "ofborg-eval-started".to_owned(),
```
cole-h, line 45: Graham has expressed the desire to remove the `grahamc` prefix from all the `ofborg` checks (https://twitter.com/grhmc/status/1353171107550027776 as a recent example), so I decided to do that for this new one here.

cole-h (approved):
Diff LGTM, thanks.

cole-h:
Note that I haven't deployed this change yet. In order to streamline this: which PR, if any, needs to be merged after the deploy succeeds?

zowoq:
> which PR, if any, needs to be merged after the deploy succeeds?

https://github.com/NixOS/nixpkgs/pull/96707

cole-h:
Done.

PR #528, opened 2020-09-06
users: add @tomberek

tomberek:
A recent set of failures require a set of re-evaluation. I'd go ahead and run them, but it seems being in the trusted list is required.

Been a part of the community since at least 2013 (https://github.com/NixOS/nixpkgs/commit/53f05a812f14b0273a24c0aae2b3db357d33cc40).


cole-h:

Cole

PR #531, opened 2020-10-05
Php update

cole-h:
Noticed that the Hydra jobset for ofborg was failing because `php72` is gone (https://hydra.nixos.org/jobset/ofborg/release#tabs-errors).

Did literally nothing to test aside from `nix-build -A ofborg.php`.

cc @andir -- should we pin php again? There is `php73` and `php74` to choose from, at the moment.

### DIFF
--- a/nix/nixpkgs.json
+++ b/nix/nixpkgs.json
@@ -1,9 +1,9 @@
 {
   "url": "https://github.com/nixos/nixpkgs-channels.git",
-  "rev": "0f5ce2fac0c726036ca69a5524c59a49e2973dd4",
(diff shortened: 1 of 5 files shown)


andir:
We probably do not have to pin it but should write a test for it. We should be able to test it without waiting for breakage in prod.

cole-h:
Isn't that what Hydra is doing for us? Or do you mean a functionality test, and not a "does it build" test?

andir:
Yeah, an actual functional test. I started writing on earlier today. Might
PR it tomorrow.

cole-h:
I believe this is superceded by https://github.com/NixOS/ofborg/pull/563.

PR #537, opened 2020-11-07
fix tests with nixUnstable

LnL7:


ofborg/src/maintainers.rs
```diff
@@ -136,13 +136,19 @@ mod tests {
     use std::process::Command;
     use std::process::Stdio;
 
+    #[cfg(target_os = "linux")]
+    static SYSTEM: &str = "x86_64-linux";
```
cole-h: Any reason to use `static` over `const`? The only difference I'm aware of is that `const` variables are replaced with their value, while `static`s are global variables (memory address). Not sure if the distinction matters.

ofborg/test-srcs/build/default.nix
```diff
@@ -1,27 +1,27 @@
 let
-  nix = import <nix/config.nix>;
-in {
+  builder = builtins.storePath <ofborg-test-bash>;
+in
+
+{
```
cole-h: IMO, this looks weird, considering `build-pr/default.nix` has `in {`.

Since you're already touching these files, maybe `nixpkgs-fmt` could be run for uniform formatting?

Would also address my nit below.

ofborg/test-srcs/build/default.nix
```diff
@@ -1,27 +1,27 @@
...
   success = derivation {
     name = "success";
     system = builtins.currentSystem;
-    builder = builtins.storePath nix.shell;
-    args = [
-      "-c"
-      "echo hi; echo ${toString builtins.currentTime} > $out" ];
+    inherit builder;
+    args = ["-c" "echo hi; echo ${toString builtins.currentTime} > $out" ];
```
cole-h: Nit (applies to all `args = [` in these files):

```suggestion
    args = [ "-c" "echo hi; echo ${toString builtins.currentTime} > $out" ];
```

or

```suggestion
    args = ["-c" "echo hi; echo ${toString builtins.currentTime} > $out"];
```

(matching spaces -- or lack thereof -- on both ends)

cole-h (commented):
"The tests that use nix don't work with nixUnstable" -> Which ones are these? Maybe they could be tagged with a `TODO` / `FIXME` to that effect.

ofborg/src/maintainers.rs
LnL7: Nope, I forgot both are a thing. 😄

ofborg/test-srcs/build/default.nix
LnL7: I'm not a big fan of nixpkgs-fmt especially since there isn't really a clear cut way to format things in certain cases. I particularly don't like some of the whitespace choices, but the expressions don't change that very often so I don't care that much.

LnL7:
@cole-h The 2 commits afterwards fix runnign the tests with nixUnstable. Everything succeeds locally again for me after this.

```
test result: ok. 83 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

ofborg/test-srcs/build/default.nix
cole-h: Fair. It doesn't have to be `nixpkgs-fmt`, as long as the formatting is (mostly) consistent :D 

(But I see you've already used it, so 🤷 )

cole-h:
Ah, got it. (As you can tell, I didn't actually test it myself hehe.)

cole-h (approved):
LGTM, and all tests pass now.

PR #538, opened 2020-11-08
config.public.json: add NickHu to trusted_users

NickHu:
On quite a few occasions, it would've been nice to be able to test my nix expressions on darwin, and I don't have access to any mac machine otherwise. In particular, right now it would be nice to test this https://github.com/NixOS/nixpkgs/pull/102763#issuecomment-723525952.


cole-h:

Cole

PR #541, opened 2020-11-25
Add lukegb to trusted users

lukegb:
It's perfectly safe, I assure you.

It's perfectly safe GIF from Alien Covenant


cole-h:
(Just FYI, this does nothing while our Darwin builder is down -- all it would do is add Darwin to the list of allowed builds.)

PR #542, opened 2020-11-25
config.public.json: add @supersandro2000

SuperSandro2000:
I would like to re-eval PRs if master was broken instead of nudging people to force push and start/build package from time to time.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -97,6 +97,7 @@
             "samueldr",
             "shlevy",
             "srhb",
+            "supersandro2000",
             "talyz",
             "teto",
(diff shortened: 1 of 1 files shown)


cole-h:
You don't need to be a trusted user to re-eval. Just `@ofborg eval` works. Same with prompting ofborg to build a package -- `@ofborg build pkgAttr`.

The only thing reserved for [trusted users](https://github.com/NixOS/ofborg#trusted-users) nowadays is building on Darwin (which doesn't matter at the moment, considering our Darwin builder is dead).

PR #544, opened 2021-01-03
typo: correct base checkout message

lukegb:
At present, when checking out the base branch, ofborg reports "Checking
original out paths", where I think it was intended to write "Checking
out original paths".


lukegb:
Staring at this more, this is actually as intended. My brain burped a bit here.

cole-h:
Maybe it would be clearer as `Checking original outPaths`.

PR #545, opened 2021-01-06
config.public.json: configure systemd topic label

alyssais:
I’ve used `nixos/modules/system/boot/systemd` to capture e.g. systemd.nix, systemd-unit-options.nix, etc. This should work as `PathTagger` just does a substring match, but I’m not sure if that’s something you’d like to depend on? If so, we’d have to move all the systemd options into a directory of their own (which might not be a bad thing anyway), or list them all out explicitly.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -215,6 +215,10 @@
         "6.topic: steam": [
             "pkgs/games/steam"
         ],
+        "6.topic: systemd": [
+            "pkgs/os-specific/linux/systemd",
+            "nixos/modules/system/boot/systemd"
+        ],
         "6.topic: TeX": [
             "pkgs/tools/typesetting/tex",
             "doc/languages-frameworks/texlive.xml"


config.public.json
```diff
@@ -215,6 +215,10 @@
         "6.topic: steam": [
             "pkgs/games/steam"
         ],
+	"6.topic: systemd": [
+            "pkgs/os-specific/linux/systemd",
+            "nixos/modules/system/boot/systemd"
+	],
```
cole-h: Alignment's slightly off.

```suggestion
        "6.topic: systemd": [
            "pkgs/os-specific/linux/systemd",
            "nixos/modules/system/boot/systemd"
        ],
```

cole-h (changes requested):
I think it's fine to rely on the substring. That said, do we want to add any of the tests to the list of monitored files?

config.public.json
SuperSandro2000: Also please do not mix tabs and spaces.

cole-h (approved):
LGTM.

~~EDIT: Waiting on Graham to fix our deployment stuff before merging (so I don't forget to deploy this afterwards).~~

cole-h:
A bit late, but everything's all good now. Thanks!

PR #546, opened 2021-01-16
tasks/evaluate: error on PRs targeting nixos-* or nixpkgs-* branches

cole-h:
A PR's target branch is easy to overlook. ofborg should complain loudly
if a nixos-* or nixpkgs-* branch is the target of a PR.

https://github.com/NixOS/nixpkgs/issues/109384, https://logs.nix.samueldr.com/nixos/2021-01-14#4470635

---

Happy to bikeshed on the actual message, but I think this is the right thing to do (and probably should have been done long ago).

EDIT: pedantry fixed here: https://github.com/NixOS/ofborg/pull/547

### DIFF
--- a/ofborg/src/tasks/evaluate.rs
+++ b/ofborg/src/tasks/evaluate.rs
@@ -334,6 +334,17 @@ impl<'a, E: stats::SysEvents + 'static> OneEval<'a, E> {
             None => String::from("master"),
         };
 
+        if target_branch.starts_with("nixos-") || target_branch.starts_with("nixpkgs-") {
+            overall_status.set_with_description(
+                "The branch you have targeted is a read-only mirror for channels. \
+                    Please target release-* or master.",
+                hubcaps::statuses::State::Error,
+            )?;
+
+            info!("PR targets a nixos-* or nixpkgs-* branch");
+            return Ok(self.actions().skip(&job));
+        };
+
         overall_status.set_with_description(
             format!("Checking out {}", &target_branch).as_ref(),
             hubcaps::statuses::State::Pending,


LnL7 (approved):
Hmm, I seem to remember that merging to channel branches didn't work. Did something change or is this perhaps something explicit per branch that wasn't updated with 20.09?

cole-h:
Don't mean to shame anyone, but the person who merged the PR to nixos-20.09 was an Owner in the org (and thus was able to bypass the protected branch restriction). It's easy to overlook, so I think it makes sense to raise an alarm, especially for cases like these.

LnL7:
Ah that explains it. Either way this is an extra place to catch it and it's something it gives feedback to the contributor. Perhaps the message could even be improved to give a bit more detail on what's going on with some of the information from eg. https://github.com/NixOS/nixpkgs/blob/master/.github/CONTRIBUTING.md#backporting-changes.

Currently it's up to the reviewer to explain https://github.com/NixOS/nixpkgs/pulls?q=base%3Anixos-20.0

ofborg/src/tasks/evaluate.rs
```diff
@@ -334,6 +334,14 @@ impl<'a, E: stats::SysEvents + 'static> OneEval<'a, E> {
             None => String::from("master"),
         };
 
+        if target_branch.starts_with("nixos-") || target_branch.starts_with("nixpkgs-") {
+            overall_status
+                .set_with_description("Improper target branch", hubcaps::statuses::State::Error)?;
```
cole-h: @LnL7 Something like this?

```suggestion
                .set_with_description("The branch you have targeted is a read-only mirror for channels. Please target release-* or master.", hubcaps::statuses::State::Error)?;
```

LnL7: Yeah, something like that is a bit more actionable.

cole-h:
Even if https://github.com/NixOS/nixpkgs/pull/109543 is accepted and merged, I'd also like to have a check in ofborg itself.

PR #550, opened 2021-02-20
config.public.json: add paths for the "6. topic: ocaml" tag

sternenseemann:
There are some out of pkgs/development/ocaml-modules paths like
pkgs/applications/networking/jackline which would be nice, but
ultimately maintaining that list would be a bit of a nightmare.

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -177,6 +177,14 @@
         "6.topic: nixos": [
             "nixos"
         ],
+        "6.topic: ocaml": [
+            "doc/languages-frameworks/ocaml.section.md",
+            "pkgs/top-level/ocaml-packages.nix",
+            "pkgs/development/ocaml-modules",
+            "pkgs/development/compilers/ocaml",
+            "pkgs/development/tools/ocaml",
+            "pkgs/development/compilers/reason"
+        ],
         "6.topic: pantheon": [
             "pkgs/desktops/pantheon",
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Would it be an option to move jackline into another directory?

sternenseemann:
No, I think this drawback is worth it actually. We want jackline etc. as top-level attributes in idiomatic places and they are not needed in the `ocamlPackages` set. Also this problem affects only a few attributes where some aren't actually clear cut topic ocaml like the haxe compiler, coq, …

Just want to make some edge cases clear in the commit message and explain my decision, really.

config.public.json
```diff
@@ -177,6 +177,13 @@
         "6.topic: nixos": [
             "nixos"
         ],
+        "6.topic: ocaml": [
+            "pkgs/top-level/ocaml-packages.nix",
+            "pkgs/development/ocaml-modules",
```
cole-h, line 183: Nit:

```suggestion
            "pkgs/development/ocaml-modules/",
```

sternenseemann: I've opted to remove all trailing slashes since this seems to be more common in the config overall.

cole-h (approved):

cole-h:
ofborg has been redeployed with this change. Thanks :)

PR #553, opened 2021-03-03
config.public.json: remove `tag_paths`

zowoq:
These labels have been moved to a github action in the nixpkgs repo.

https://github.com/NixOS/nixpkgs/pull/113077


config.public.json
```diff
@@ -114,158 +114,7 @@
...
-        ],
-        "8.has: documentation": [
-            "doc",
-            "nixos/doc"
-        ],
-        "8.has: module (update)": [
-            "nixos/modules"
-        ]
-    },
+    "tag_paths": {},
```
cole-h: Rather than just emptying the list, could you also remove all `tag_paths` related code? There's no point in keeping around dead code.

If you don't have the time / aren't comfortable with this, I could do it before I merge. Whatever works best for you; just let me know.

cole-h (changes requested):

config.public.json
zowoq: Done. Tests pass so I'm hoping it's fine.

cole-h (approved):
Diff LGTM. Starting the deploy process soon. Once successful, I'll merge the Nixpkgs PR.

PR #556, opened 2021-04-18
config.public.json: configure bsd topic label

alyssais:


alyssais:
@SuperSandro2000 has informed me that this has moved to https://github.com/NixOS/nixpkgs/blob/master/.github/labeler.yml.

PR #562, opened 2021-05-16
WIP: Introduce an one-off-build command to reproduce ofBorg locally

andir:
Motivated by trying to review a change in how ofBorg reports build
errors I started hacking on a command that allowed me to execute just
the build part of ofBorg without setting up a local queue or config
files.

This is still a WIP that is hardcoded to exactly one PR and commit
revision. I plan to convert all the constants to Clap CLI arguments once
I moved the build system of the project to naersk and can then introduce
new dependencies.

Things to be done:

 - [ ] Move constants to Clap
   - [ ] The system attribute should be defaulted to the current system.
         We can use the `target_os` and `target_arch` values for that.
	 The user should still be able to override thos tho...
 - [ ] Make the logging a bit prettier
 - [ ] Implement a custom Notification receiver that just logs the
       messages as they come in.
 - [ ] (idea) add support for the eval part as well so we can run ofBorg
        locally just like we can run nixpkgs-review.

### DIFF
--- a/ofborg/src/bin/one-off-builder.rs
+++ b/ofborg/src/bin/one-off-builder.rs
@@ -1,0 +1,70 @@
+/// Build attributes within a Nixpkgs PR the way that ofBorg does
+use ofborg::nix::Nix;
+use ofborg::notifyworker::SimpleNotifyWorker;
+use ofborg::tasks::build::BuildWorker;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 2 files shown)


wolfgangwalther:
Today, Nixpkgs GHA workflows can mostly be reproduced by `nix-build ci`, eval be run by `nix-build ci -A eval.full` and packages be build via nixpkgs-review.

PR #564, opened 2021-05-17
Disable trusted users for now

### LINKED ISSUE #559: Disable trusted users by default
The current Darwin build machine is restarted and erased relatively frequently, so the sandboxing isn't as much of a worry anymore

- [x] Add `"enable_trusted_users": true/false,` toggle to `config.public.json`
- [x] Skip ACL stuff if that is `false`
- [x] Update README
- [x] Close all trusted user PRs

We don't want to remove the ACL stuff, just in case... It's easy to toggle back on if anything unexpected happens.

cole-h:
The current darwwin builder is reset very, very frequently (mostly due to its
storage constraints necessitating it), so there's much less of a reason to limit
the people who can utilize it. (Enabling it for everybody will also guarantee
more frequent resets, as well.)

However, it is kept as an option so that it can be re-enabled some time in the
future, if anything were to happen.

---

Closes https://github.com/NixOS/ofborg/issues/559.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,11 @@ combinations:
     @ofborg build list of attrs looks good to me!
     ```
 
-## Trusted Users
+## Trusted Users (Currently Disabled)
+
+> **NOTE:** The Trusted Users functionality is currently disabled, as the
+> current darwin builder is reset very frequently. This means that _all_ users
+> will have their PRs build on the darwin machine.
 
 Trusted users have their builds and tests executed on _all_ available platforms,
 including those without good sandboxing. Because this exposes the host to a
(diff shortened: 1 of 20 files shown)


grahamc (approved):

README.md
```diff
@@ -122,7 +122,11 @@ combinations:
     @ofborg build list of attrs looks good to me!
     ```
 
-## Trusted Users
+## Trusted Users (Currently Disabled)
+
+> **NOTE:** The Trusted Users functionality is currently disabled, as the
+> current darwin builder is reset very frequently. This means that _all_ users
+> can have their PRs build on the darwin machine.
```
grahamc: `will have`

grahamc (approved):

grahamc (approved):

PR #572, opened 2021-07-16
check unfree packages for deprecated aliases

SuperSandro2000:
This is probably causing some regressions but after removing unfree aliases from https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/office/timedoctor/default.nix I think it is a good idea to do that.

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -435,7 +435,7 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
                     String::from("."),
                     String::from("--arg"),
                     String::from("config"),
-                    String::from("{ allowAliases = false; }"),
+                    String::from("{ allowAliases = false; allowUnfree = true; }"),
                 ],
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
@cole-h could you take a look at this?

ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -435,7 +435,7 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
                     String::from("."),
                     String::from("--arg"),
                     String::from("config"),
-                    String::from("{ allowAliases = false; }"),
+                    String::from("{ allowAliases = false; allowUnfree = true; }"),
```
SuperSandro2000, line 438:
```suggestion
                    String::from("{ allowAliases = false; allowBroken = true; allowUnfree = true; }"),
```

SuperSandro2000:
@cole-h can you take a look at this PR?

cole-h:
https://github.com/NixOS/ofborg/pull/645 has been merged, which achieves a similar goal.

PR #573, opened 2021-08-02
Label PRs with "bsd" in the title

alyssais:
I haven't tested this beyond `cargo test`, because I don't really know how.


r-burns:
Should we try to do something clever with word splitting so it doesn't match on e.g. "libsdl"?

alyssais:
That would be better for sure, but exceeds the amount of energy I can put into this, since it would be changing quite a lot how this code works.

ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -26,6 +26,12 @@ use uuid::Uuid;
 
 static MAINTAINER_REVIEW_MAX_CHANGED_PATHS: usize = 64;
 
+const TITLE_LABELS: [(&str, &str); 3] = [
+    ("darwin", "6.topic: darwin"),
+    ("macos", "6.topic: macos"),
```
cole-h: Nit, but we don't have a macos label :P

```suggestion
    ("macos", "6.topic: darwin"),
```

cole-h (commented):
I'd be happy to get this in in its current form (once my singular nit is addressed), even with the potential for false-positive labels. As unlikely as it is now, if somebody added a project by the name of `darwini` or `sigmacosine`, those would also be falsely tagged. As long as you're fine with the potential for noise in this label, this can be addressed later (when I or another interested party has time).

ofborg/src/tasks/eval/nixpkgs.rs
alyssais: Fixed!  I also alphabetised the list because that seemed like a sensible thing to do.

cole-h:
I've let this grow long in the tooth, and I'm sorry about that.

I just pushed a commit that uses the `regex` crate so that we can match on word boundaries using the `\b` anchor. This is slightly more robust than just checking for the existence of the string "bsd" or "macos" or "darwin" in the title. (`regex` was already in our dependency tree, so it's almost free using it :)

I hope to merge this sometime soon (but for real this time).

cole-h (approved):
Thanks again :)

PR #574, opened 2021-08-02
Don't use global git config in tests

alyssais:
The tests were failing for me because I have git configured to use
"main" instead of "master" as the default initial branch name for new
repositories.  Having the tests use global git configuration from
developer systems is just asking for trouble with non-reproducibility,
but fortunately git gives us a mechanism to disable it.

This also means we don't need to worry about GPG any more, and while
I'm here I've also used environment variables to configure the author
name and email so it doesn't have to be specified every time.


cole-h (approved):
Thanks (and sorry for the wait)!

PR #579, opened 2021-10-01
Never request reviews on staging-esque branches

sternenseemann:
PRs from staging-next usually never cause ofborg to request reviews from
package maintainers because the change count will be high enough even in
the beginning to prevent it. However the key insight is that we don't
want to not request reviews because there would be too many, but rather
the changes have already been reviewed (before being merged into staging
for example).

There are in fact branches similar to staging-next, like
haskell-updates, where ofborg will still request reviews, but
unnecessarily so, because the branch mostly contains already reviewed
and merged PRs. In https://github.com/NixOS/nixpkgs/pull/139967 in
particular you can see a good example of this, including Damien Cassou
fighting the ofborg request for review…

This commit adds a new condition to the review requesting logic and
checks if the head of the PR is a staging branch (i. e. matches
staging-next* or haskell-updates and is in the NixOS/nixpkgs
repository). This should prevent such cases reliably in the future.
The list of branches can be extended easily.

---

Help with testing / reviewing would be appreciated, I can't really run ofborg locally to test this change…

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -252,6 +252,24 @@ impl<'a> NixpkgsStrategy<'a> {
 
             let prefix = get_prefix(self.repo.statuses(), &self.job.pr.head_sha)?;
 
+            let staging_branch = self.pull
+                .get()
+                .map(|p| is_staging_branch(&p.head))
+                .unwrap_or(false);
+
+            if staging_branch {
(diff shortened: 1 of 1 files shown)


Profpatsch:
Bump

wolfgangwalther:
This has been implemented in nixpkgs' GHA workflows in https://github.com/NixOS/nixpkgs/pull/418496.

It doesn't apply to codeowners, yet, but that's a purely GHA issue anyway, those were never managed by ofborg.

PR #583, opened 2021-11-11
Support aarch64-darwin

grahamc:


ofborg/src/outpaths.nix
```diff
@@ -11,6 +11,7 @@ let
     {
       supportedSystems = [
         "aarch64-linux"
+        # "aarch64-darwin" # !!!
```
grahamc, line 14: Adding this might cause significant problems around memory utilization, so I'm hesitant to add it in this same PR.

cole-h (approved):
LGTM. CI fix is in https://github.com/NixOS/ofborg/pull/584.

ofborg/src/outpaths.nix
andir: I wonder if eval shouldn't do one system at a time instead of as one big expression.

cole-h:
:crossed_fingers: Let's give this a shot.

ofborg/src/outpaths.nix
Mic92: In that case I need to special case it in nixpkgs-review since I rely on ofborg's eval output to compute which packages needs a rebuild

PR #592, opened 2022-02-21
Support building x86_64 packages on aarch64-darwin

### LINKED ISSUE #591: ofborg should be able to treat aarch64-darwin as x86_64-darwin for builds
The M1s have Rosetta 2, which we can leverage to get extra builds going for x86_64-darwin.

This is inspired by the fact that we had only one x86_64-darwin builder for a while -- one of the x86_64 machines is currently unavailable, one was stuck on a build for over a month, somehow, and one has stopped responding to pings / SSH altogether.

One idea I've had is to just special-case aarch64-darwin in the builder binary to also subscribe to the x86_64-darwin build queue. This is the ideal path, because it (likely) means that ofborg won't be trying to do 2 builds at the same time and potentially exhausting resources. If that falls through, another idea could be to just spawn another builder with a config claiming that it's x86_64-darwin.

cole-h:
Closes https://github.com/NixOS/ofborg/issues/591.

---

Needed to bump and add the `unstable` feature to async-std for access to the `join` method. Maybe there's a better way to do this, but I'm not well-versed (or really, versed at all) in Rust async.

---

<del>Going to be testing on one of the M1 machines we have access to ~soon.</del>

EDIT: Didn't work -- likely need to change the `supportedSystems` stuff in `nix.rs`. Probably also want to run the ofborg binaries manually, to see the logging output first-hand...


ofborg/src/bin/builder.rs
```diff
@@ -31,8 +29,36 @@ fn main() -> Result<(), Box<dyn Error>> {
     };
 
     let conn = easylapin::from_config(&cfg.rabbitmq)?;
+    let handle = self::create_handle(&conn, &cfg, None)?;
+
+    // aarch64-darwin can build x86_64-darwin thanks to Rosetta 2
+    if cfg.nix.system == "aarch64-darwin" {
```
grahamc: Is there a way we could make this a property of the configuration and not an automatic "upgrade"?

grahamc: For example, making the system field be a list.

cole-h: Yep, very easily (aside from bikeshedding on the config name and value). This was just the "easiest" way to do so while hacking on a single machine.

cole-h:
I don't know if this will work in the end :( Needs more investigation, but one example failure is the following: https://logs.nix.ci/?key=nixos/nixpkgs.161527&attempt_id=8ca7bc24-b413-4b61-a406-ebae2834f506. It was processed on the M1 I was testing with, but it appears there was some issue with an emulated instruction.

Requires more investigation as I have time... But it appears Rosetta 2 isn't perfect. Also, the couple M1s we have appear to be running an older version of macOS (11.0.1 at time of writing). https://github.com/golang/go/issues/42700 appears to be the problem from that log, and it's claimed that this issue was fixed in a later version of macOS (11.2).

<details><summary>Relevant log details for posterity</summary>

```
building '/nix/store/rj5ywpiqg7hn8mmgflg8n6kyahrr598s-source.drv'...
trying https://github.com/redhat-developer/odo/archive/v2.5.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   138  100   138    0     0    386      0 --:--:-- --:--:-- --:--:--   386
100 27.8M    0 27.8M    0     0  6532k      0 --:--:--  0:00:04 --:--:-- 7386k
unpacking source archive /private/tmp/nix-build-source.drv-0/v2.5.0.tar.gz
building '/nix/store/z3ymirw538bkdp2a2v8vjnn19grmq2zg-odo-2.5.0.drv'...
unpacking sources
unpacking source archive /nix/store/vk7njlsnaqmqjm7m5lvgi6dck6dd3diy-source
source root is source
patching sources
configuring
building
go build -mod=vendor -ldflags="-X github.com/redhat-developer/odo/pkg/version.GITCOMMIT=" cmd/odo/odo.go
assertion failed [abi_info.kind == AbiKind::TranslatedCode]: emulated forward to an arm pc that isn't in translated code. arm_pc=0x101cb44d8 abi_kind=6 emulation_interval=[0x101d2af28,0x101d2af3c) instruction_interval=[0x101d2af14, 0x101d2af3c) x86_rip=0x10d28e2
(ThreadContextRegisterState.cpp:677 move_to_instruction_boundary)
error: building of '/nix/store/z3ymirw538bkdp2a2v8vjnn19grmq2zg-odo-2.5.0.drv!out' from .drv file timed out after 3600 seconds
error: 1 dependencies of derivation '/nix/store/6ixl053g0bav4kdjjigi13p88n0gh4j1-odo-2.5.0-test-version.drv' failed to build
error: build of '/nix/store/6ixl053g0bav4kdjjigi13p88n0gh4j1-odo-2.5.0-test-version.drv', '/nix/store/z3ymirw538bkdp2a2v8vjnn19grmq2zg-odo-2.5.0.drv' failed
```

</details>

ofborg/src/bin/builder.rs
cole-h: I added `additional_build_systems` for this purpose -- modifying `systems` to become a list would be somewhat of a pain due to the fact that it's used in the `whoami()` string. We could drop the `system` from the string, or we could stick with this `additional_build_systems`...

ofborg/src/config.rs
```diff
@@ -41,6 +41,7 @@ pub struct RabbitMqConfig {
 #[derive(Serialize, Deserialize, Debug)]
 pub struct NixConfig {
     pub system: String,
+    pub additional_build_systems: Option<Vec<String>>,
```
grahamc: What would happen if we changed `system` to `Vec<String>`?

cole-h: It would require deeper-reaching changes because `system` is also used as part of the `whoami()` string. We could just pick the first system from the list as the `whoami`, but at this point it's not possible to choose between the two based on anything.

grahamc: Cool, I think that is actually a useful thing to put into the whoami. Maybe something like this:

```
    let systems = vec![String::from("x86_64-darwin"), String::from("aarch64-darwin")];
  let whoami = systems.join(",");
```

(for example.)

cole-h:
Note that this has only been deployed to 1 of our M1s -- the other still needs to be updated (both macOS and the config file).

cole-h:
Both M1s are now running this change :tada:

PR #593, opened 2022-03-07
tasks/eval/nixpkgs: don't re-request reviews

ncfavier:
Don't request a new review from a maintainer who has already reviewed a PR.

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -568,12 +568,25 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
 fn request_reviews(maint: &maintainers::ImpactedMaintainers, pull: &hubcaps::pulls::PullRequest) {
     let pull_meta = pull.get();
 
+    let pull_requested_reviews = pull.review_requests().get();
+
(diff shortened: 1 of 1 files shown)


cole-h:
Is there any particular reason _why_ you don't want a new review requested after changes have been made?

ncfavier:
Reviewers are subscribed to the PR so they were already notified of the changes. I think it should be up to the PR author to decide whom and when to rerequest a review from.

But the motivating case that just happened to me is: maintainer adds a review before ofborg finishes eval, then ofborg wakes up and requests a review from them again.

SuperSandro2000:
With that feature you can easily filter notifications for more important ones https://github.com/notifications?query=reason%3Areview-requested

If I am tight on time I am not reacting to the 100+ nixpkgs notifications I get per day but the review rerequest shows me that someone pushed to the PR and most of the time has fixed my remarks.

ncfavier:
Ok, that's fair. Maybe we can skip rerequests only on the first ofborg eval, then.

SuperSandro2000:
I don't know how easy that is to do and it sounds like quite a complication for such a small thing in my opinion.

ncfavier:
Oh well.

PR #594, opened 2022-04-01
always disallow aliases

### LINKED ISSUE #575: Always evaluate with `config.allowAliases = false` to avoid deprecation warning errors
Since #250, we've been checking that Nixpkgs doesn't depend on aliases.

In https://github.com/NixOS/nixpkgs/pull/134230 I've had to revert deprecation warnings in order to let ofborg evaluate cleanly.

If we apply `allowAliases = false` to all evaluations, we will be able to provide helpful deprecation warnings without causing ofborg to fail.

It will also save an evaluation, because the separate check from #250 will be implicit in the other evals.

Artturin:
remove package-list-no-aliases
this will also save us an eval

Closes https://github.com/NixOS/ofborg/issues/575

Supersedes https://github.com/NixOS/ofborg/pull/572 https://github.com/NixOS/ofborg/pull/539

### DIFF
--- a/ofborg/src/outpaths.nix
+++ b/ofborg/src/outpaths.nix
@@ -19,6 +19,7 @@ let
 
       nixpkgsArgs = {
         config = {
+          allowAliases = false;
           allowBroken = true;
           allowUnfree = true;
           allowInsecurePredicate = x: true;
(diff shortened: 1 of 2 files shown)


Artturin:
one issue with this may be that no ci will eval the aliases file?

Artturin:
also is this correct or does the allowAliases config need to be passed to all the eval checks?

Artturin:
making a another pr which does this better and will still catch errors in aliases.nix

PR #596, opened 2022-04-02
buildresult.rs: use skipped instead of neutral for a skipped job

Artturin:
doesn't make much sense for a skipped job to be neutral


Artturin:
Ping

Artturin:
there was a release to https://github.com/cachix/install-nix-action/releases/tag/v17

PR #597, opened 2022-04-20
disallow aliases in the nixpkgs-tarball eval

Artturin:
this allows ofborg to still check the aliases.nix file for errors
unlike https://github.com/NixOS/ofborg/pull/594 which would have
disabled aliases in outpaths.nix

errors in the alias.nix file are caught by outpaths.nix eval so we can
disable aliases here


cole-h (approved):
Let's give it a shot.

cole-h:
This doesn't appear to work. `pkgs/top-level/release.nix` does not take a `config` argument, but a `nixpkgsArgs` argument which may contain `config`. However, even then, this wouldn't work -- these errors don't appear at eval time, but build time.

Compare the following running off of https://github.com/NixOS/nixpkgs/pull/169822:

```
nix-instantiate  --arg nixpkgs '{ outPath = ./.; revCount = 1; shortRev = "73382eb"; rev = "73382ebb720506d43dd0ef031d5cb19a1b96f678"; }' --arg nixpkgsArgs '{ config = { allowAliases = false; }; }' ./pkgs/top-level/release.nix -A tarball
# succeeds
```

and

```
nix-build --arg nixpkgs '{ outPath = ./.; revCount = 1; shortRev = "73382eb"; rev = "73382ebb720506d43dd0ef031d5cb19a1b96f678"; }' --arg nixpkgsArgs '{ config = { allowAliases = false; }; }' ./pkgs/top-level/release.nix -A tarball
# fails due to alias
```

cole-h:
Reverted in #598.

This may be helpful -- the error during build comes after the following:

```
++ nix-env -f /nix/store/7haipzvw641rgsfak0w9g57xy6nj9qb8-master --show-trace --argstr system x86_64-linux --arg config '{ allowAliases = false; }' --option experimental-features no-url-literals -qa --drv-path --system-filter '*' --system --option build-users-group ''                                                                                    
error: undefined variable 'dateutil'                                                                                                                                            
```

So maybe the necessary change is above in the `list-packages-no-aliases` step (roughly line 451)?

PR #609, opened 2022-08-12
tasks/eval/nixpkgs: build the nixos manual / options to verify them

cole-h:
Just evaluating them isn't enough anymore, now that the docs are lazy.

---

Maybe we should keep the instantiate step but add the build step? idk

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -485,7 +485,7 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
             ),
             EvalChecker::new(
                 "nixos-options",
-                nix::Operation::Instantiate,
+                nix::Operation::Build,
                 vec![
                     String::from("--arg"),
                     String::from("nixpkgs"),
@@ -498,7 +498,7 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
             ),
             EvalChecker::new(
                 "nixos-manual",
-                nix::Operation::Instantiate,
+                nix::Operation::Build,
                 vec![
                     String::from("--arg"),
                     String::from("nixpkgs"),


cole-h:
There's a NixOS option that will allow us to go back to the old way and keep checking at eval-time: `documentation.nixos.options.splitBuild = false` (thanks pennae!)

Now just need to set that somehow.

Artturin:
> There's a NixOS option that will allow us to go back to the old way and keep checking at eval-time: `documentation.nixos.options.splitBuild = false` (thanks pennae!)
> 
> Now just need to set that somehow.

maybe this but the drv still contains lazy-options.json
```
nix-instantiate --arg configuration '{ ... }: { documentation.nixos.options.splitBuild = false; }' ./nixos/release.nix -A options
```

ncfavier:
Is this really needed? The GitHub [workflow](https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/manual-nixos.yml) already builds the NixOS manual whenever a file in `nixos/` is modified.

cole-h:
Well, we have two options: 1) make this actually build the manual / don't be lazy when instantiating it; or 2) disable the functionality in ofborg, since we _do_ have GHA setup for that. This is assuming the GHA workflow does indeed catch issues with the manual (I don't know).

ncfavier:
> This is assuming the GHA workflow does indeed catch issues with the manual

It does.

I think we can go with 2, unless there's a need to do this in ofborg? GHA are a bit nicer, and they're already working.

lheckemann (approved):
Given the discussion linked above (GHA doesn't run on backport PRs), I think it would be good to introduce this if it's feasible.

wolfgangwalther:
The nixpks and nixos manuals are now built in nixpkgs' GHA workflows. Both in PRs targeting stable and unstable branches.

PR #611, opened 2022-08-18
config.public.json: bump build_timeout_seconds to 3 hours from 1 hour

Artturin:
the most common isue i see on darwin builds is 'timed out after 3600 seconds' due to llvm or gcc failing to build which makes the darwin ofborg useless on many staging PR's

for example
https://logs.nix.ci/?key=nixos/nixpkgs.185972&attempt_id=a7f34cd2-8bce-4299-a4e5-d6bc6f5c434f
is being built with -j1

3 hours is picked from the time it took to build 'nixpkgs:staging-next:gcc-unwrapped.x86_64-darwin' 2h 45m 16s
https://hydra.nixos.org/build/187709773

### DIFF
--- a/config.public.json
+++ b/config.public.json
@@ -121,7 +121,7 @@
     "nix": {
         "system": "x86_64-linux",
         "remote": "daemon",
-        "build_timeout_seconds": 3600,
+        "build_timeout_seconds": 10800,
         "initial_heap_size": "4g"
     }
 }


grahamc (commented):
I wonder how this will impact other PRs, and if we'll cause a giant backlog of builds that take forever. The goal of ofborg hasn't necessarily been to build every PR, but to provide decent coverage for all PRs.

That said, I'm open to changing it. However if it causes a backlog we may have to go back.

Artturin:
👍let's try it

cole-h:
I've applied this to (almost) all of the darwin machines. I'll check back in tomorrow to see if it creates a giant backlog.

I applied this without merging because the config for the darwin machines are all dropped onto the boxes by hand, unfortunately.

I'll leave this open until we decide if it's fine, or it causes too many problems.

cole-h:
This has created a giant backlog. I'm going to revert this back to 1 hour.

PR #615, opened 2022-09-22
clone: attempt to clean git if checking out fails

cole-h:
Somehow, some evaluators have had dirty git checkouts, causing them to continually crash. We can fix this by attempting to cleanup the checkout if we fail to switch to a different commit. If checking out another commit still fails after that, then it is acceptable to crash.


SuperSandro2000:
Why are we not using ``git checkout -f`` here? I don't think we care about the dirty state and should happily overwrite all changes.

Artturin:
#625

PR #563, opened 2021-05-16
Update nixpkgs & cleanup shell.nix

andir:
This PR updates nixpkgs to the latest channel revision of nixos-unstable and also cleans up a bit of the craft that accumulated in the shell.nix over the years. It also fixes / silences a few of the lints that have been introduced by newer versions of clippy that nobody has looked at but that caused CI failures.

PR #618, opened 2022-10-17
update cargo deps

Artturin:


cole-h:
Could you please explain why this is necessary?

Artturin:
> Could you please explain why this is necessary?

It's good to keep dependencies up to date (security, bugfixes)

it is nicer to develop when dependencies are up to date

if dependencies aren't updated often then the program could get stuck on old dependencies due to no one willing to update them

cole-h:
If you move the PHP updates to another PR, I'd be happy to take a look at this one. I know next to nothing about PHP, and even less about how the PHP stuff integrates with ofborg, so I'd rather be able to easily revert that PR separately from this one if anything were to go wrong.

php/default.nix
```diff
@@ -1,11 +1,11 @@
 {pkgs ? import <nixpkgs> {
     inherit system;
-  }, system ? builtins.currentSystem, noDev ? false}:
+  }, system ? builtins.currentSystem, noDev ? false, php ? pkgs.php, phpPackages ? pkgs.phpPackages}:
 
 let
   composerEnv = import ./composer-env.nix {
-    inherit (pkgs) stdenv lib writeTextFile fetchurl unzip phpPackages;
-    php = pkgs.php74;
```
Artturin: @cole-h there's no php <8 available in nixpkgs, that's why I upgraded it and the dependencies like done in https://github.com/NixOS/ofborg/pull/563

Ping @grahamc because it looks like he wrote the code

Artturin: i ran `php -l` (syntax check) on the php files and no issues

Artturin:
> If you move the PHP updates to another PR, I'd be happy to take a look at this one. I know next to nothing about PHP, and even less about how the PHP stuff integrates with ofborg, so I'd rather be able to easily revert that PR separately from this one if anything were to go wrong.

only cargo update here now

Artturin:
#622

PR #630, opened 2023-01-02
use `nixosTests` for tests

### LINKED ISSUE #504: support test *sets*
When you specify an attrset containing tests instead of concrete tests (possibly by mistake), you only get a confusing error message and it succeeds (!):
```
warning: ignoring the user-specified setting 'restrict-eval', because it is a restricted setting and you are not a trusted user
```

Perhaps it wouldn't be too hard to run all the tests in the set.  From tests already merged usage could be e.g. `@ofborg test boot`

_[A real-life occurrence.](https://github.com/NixOS/nixpkgs/pull/76542#issuecomment-633657274)_

ncfavier:
Fixes https://github.com/NixOS/ofborg/issues/504? Surely it can't be that simple. cc @cole-h

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ instructions for ofborg.
 @ofborg test list of tests
 ```
 
-This will run `nix-build ./nixos/release.nix -A tests.list -A tests.of -A
-tests.tests` from the root of the Nixpkgs checkout.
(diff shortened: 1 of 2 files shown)


cole-h (approved):
Let's find out!

PR #638, opened 2023-03-27
log_message_collector: open log handle to ensure log file exists 

cole-h:
In the past, we've seen some PRs have empty build logs, which would display a 404 if nothing was ever written to its log output. This should hopefully prevent that from happening.


cole-h:
I added a test so I'm sure it'll be fine.

Surely.

PR #640, opened 2023-04-12
Add support for building only on a single system

### LINKED ISSUE #232: Feature request: build only for specific platform
Would be great if we could ask ofborg to build a package only for a specific platform, like
`@GrahamcOfBorg build foo.aarch64-linux`. 

This can significantly save resources (and reduce noise on Github) when debugging PRs.

hesiod:
This PR adds support for building only on a single system. The proposed syntax based on `build_system` is only a suggestion, I'm open to better ideas.

I took the liberty of breaking the comment parser into a few modular sub-parsers in the process, as it was a little convoluted previously. In any case, the parser is still using the relatively old `nom` version 4.2.3. Newer nom versions use functions instead of macros (see the [migration guide](https://github.com/rust-bakery/nom/blob/main/doc/upgrading_to_nom_5.md#from-macros-to-functions)). At some point the parser should be rewritten using this new style, but I didn't do it in this PR to keep it simple.

Note: I can't test the PR using a live builder as I'm not running one. I'm opening the PR anyway as the README says this is OK due to the lack of test environments for ofborg.

Closes #232.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,19 @@ the Nixpkgs checkout (see also "[How does ofborg call
 Builds will run on all allowed machines. For more information, see the "[Trusted
 Users](#trusted-users)" section.
 
+### build_system
+
+```
+@ofborg build_system SYSTEM list of attrs
+```
+
+Same as [build](#build), but restricts building to only one platform specification
+instead of attempting to build on all allowed and supported systems.
+This can be helpful to reduce needless builds and noise when debugging a platform-specific issue.
+
+Only the currently supported Nixpkgs systems can be passed to `build_system`:
+`x86_64-linux`, `aarch64-linux`, `x86_64-darwin`, and `aarch64-darwin`.
+
 ## Multiple Commands
 
 You can use multiple commands in a variety ways. Here are some valid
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -81,6 +81,19 @@ the Nixpkgs checkout (see also "[How does ofborg call
...
 
+### build_system
+
+```
+@ofborg build_system SYSTEM list of attrs
+```
+
+Same as [build](#build), but restricts building to only one platform specification
+instead of attempting to build on all allowed and supported systems.
+This can be helpful to reduce needless builds and noise when debugging a platform-specific issue.
```
SuperSandro2000, line 92: The need for this roots from a different problem: people using ofborg for debugging which is not great. For every push the eval task runs which is rather expensive.

hesiod: I'm a little conflicted about this. Certainly ofborg should not be the primary development tool while working on a PR. On the other hand, if you're trying to debug a build failure on a foreign architecture ofborg can be the only way to properly test a PR.
According to the docs and the code, it should be possible to skip evaluation by assigning the work-in-progress label to a PR. However, for some reason this does not seem to stop ofborg from running the eval task (see `ofborg/src/tasks/evaluate.rs`), at least when I tested it. Perhaps it would make sense to skip the eval task for all draft PRs.

Personally, I don't have access to any macOS systems and other than ofborg I don't know about another way to test whether a package builds on Darwin.

PR #644, opened 2023-06-10
nix: build unfree packages

### LINKED ISSUE #194: Build unfree packages
I understand the complexities of *distributing* software with an unfree license but any chance ofborg can start building those anyway?

rhendric:
Closes #194.

Example PR where this would have been helpful: https://github.com/NixOS/nixpkgs/pull/236695

### DIFF
--- a/ofborg/src/nix.rs
+++ b/ofborg/src/nix.rs
@@ -357,6 +357,7 @@ impl Nix {
             &format!("{}", self.build_timeout),
         ]);
         command.args(&["--argstr", "system", &self.system]);
+        command.args(&["--arg", "config", "{ allowUnfree = true; }"]);
 
         if self.limit_supported_systems {
             command.args(&[
@@ -693,7 +694,15 @@ mod tests {
 
         let ret: Result<fs::File, fs::File> = nix.run(command, true);
 
-        assert_run(ret, Expect::Pass, vec!["./default.nix", "-A foo -A bar"]);
+        assert_run(
+            ret,
+            Expect::Pass,
+            vec![
+                "./default.nix",
+                "-A foo -A bar",
+                "--arg config { allowUnfree = true; }",
+            ],
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
unfree packages are for a reason unfree and most likely fail to build because ofborg cannot obtain the source or we are not allowed to build the packages.

rhendric:
Wish that had been feedback left on #194. It's hard to pick off small contributions to make when issues are met with silence and only after one does the work do the thumbs-down come out. I mean, unless you don't like growing the set of contributors, in which case great job, keep it up.

On this issue, though: if the source isn't available, the build fails almost immediately. Why is that worse than failing immediately regardless of whether the source is available? Hydra builds non-free packages (I think? I assume that's where they come from when I install them and they aren't built locally); why would ofborg not be allowed to?

SuperSandro2000:
> Hydra builds non-free packages

No it doesn't because most of the time the restrictive and unfree licenses do not allow that. If hydra would build them ofcourse ofborg could, too.

JulienMalka:
I don't see how unfree packages could not be buildable at all and still distributed through nixpkgs.

SuperSandro2000:
We are not distributing them, we are distributing code to be able to build them if you confirming to their licenses.

JulienMalka:
I know this, it doesn't mean we could not have ofborg build them without distributing the results, that is I believe the proposition here. I don't have any strong feeling about it, but I feel like if we are to discard this proposition, let us discard it with legit points.

SomeoneSerge:
> unfree packages are for a reason unfree and most likely fail to build because ofborg cannot obtain the source or we are not allowed to build the packages.

Isn't it exceedingly rare that we are not allowed to _build_ the derivations? Redistribution is usually the point of contention, never evaluation, and (I believe) rarely the "building". We can literally count these things by checking how many attributes with `!meta.license.free` there are, and how many derivations rely on `requireFile`

Issue #577, opened 2021-09-16
python3Packages are not built automatically


figsoda:
for example, https://github.com/NixOS/nixpkgs/pull/138049
`python3Packages.surepy` is mentioned in the commit message but not automatically built by ofborg

figsoda:
i'm not familiar with ofborg code base but looks like https://github.com/NixOS/ofborg/blob/517a32a84dd2e3f7445fb38dd64e1e6270eb350b/ofborg/src/tasks/eval/nixpkgs.rs#L315 is causing the issue

figsoda:
I think `ofborg/src/outpaths.nix` evaluates changed paths to either `python38Packages.*` or `python39Packages.*`, and `python3Packages.*` gets filtered out because it doesn't contain it

Artturin:
if we apply
```nix
diff --git a/pkgs/top-level/all-packages.nix b/pkgs/top-level/all-packages.nix
index d13066e27f3..f748a88a5b5 100644
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -14052,6 +14052,11 @@ with pkgs;
   # When switching these sets, please update docs at ../../doc/languages-frameworks/python.md
   python2 = python27;
   python3 = python39;
+
+  # pythonPackages further below, but assigned here because they need to be in sync
+  python2Packages = python27Packages;
+  python3Packages = python39Packages;
+
   pypy = pypy2;
   pypy2 = pypy27;
   pypy3 = pypy38;
@@ -14093,10 +14098,6 @@ with pkgs;
     x11Support = true;
   };
 
-  # pythonPackages further below, but assigned here because they need to be in sync
-  python2Packages = python2.pkgs;
-  python3Packages = python3.pkgs;
-
   pythonInterpreters = callPackage ./../development/interpreters/python { };
   inherit (pythonInterpreters) python27 python37 python38 python39 python310 python311 python3Minimal pypy27 pypy38 pypy37 rustpython;
 
```
to nixpkgs

`./maintainers/scripts/rebuild-amount.sh  --print HEAD`
then shows them like this
```bash
python3Packages.aadict.x86_64-darwin                                                      /nix/store/n79wdn1fyggr6gckbaabrlscd21hnq0d-python3.9-aadict-0.2.3
[...]

PR #646, opened 2023-06-15
replace python3Packages with python310Packages

Artturin:
so when its in commit message build it

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -639,6 +639,13 @@ fn parse_commit_messages(messages: &[String]) -> Vec<String> {
             let pkgs: Vec<&str> = line.split(',').collect();
             pkgs
         })
+        .map(|pkg| {
+            if pkg.contains("python3Packages") {
+                pkg.replace("python3Packages", "python310Packages")
+            } else {
+                pkg.to_owned()
+            }
+        })
         .map(|line| line.trim().to_owned())
         .collect()
 }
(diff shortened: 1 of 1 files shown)


ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -34,6 +34,10 @@ const TITLE_LABELS: [(&str, &str); 4] = [
     ("cross", "6.topic: cross-compilation"),
 ];
 
+const TITLE_REPLACEMENTS: [(&str, &str); 1] = [
+    ("python3Packages", "python310Packages"),
```
mweinelt, line 38:
```suggestion
    ("python3Packages", "python3.pkgs"),
```

That is less of a moving target and references the default python version.

Artturin: Won't work for the same reason why python3Packages won't work https://gist.github.com/GrahamcOfBorg/f9eea356a93c92631f3d2b4304070282

mweinelt: So we need to bump the version here every year, when we move the default python version? Will be 3.11 in a few weeks.

rhendric:
Fixes #577, if merged?

(I still have my probably-stupid question unanswered over there.)

ofborg/src/tasks/eval/nixpkgs.rs
mweinelt: Is python311Packages now on `python-updates` and soon on `staging`.

wolfgangwalther:
This seems to work by now.

PR #649, opened 2023-08-18
eval: Only support Nixpkgs

infinisil:
Ofborg seems only used on Nixpkgs with no intention of using it for other repos, so let's not have code to support that unless needed.


ofborg/src/tasks/evaluate.rs
```diff
@@ -290,7 +290,13 @@ impl<'a, E: stats::SysEvents + 'static> OneEval<'a, E> {
             }
         };
 
-        let mut evaluation_strategy: Box<dyn eval::EvaluationStrategy> = if job.is_nixpkgs() {
+        if !job.is_nixpkgs() {
+            error!(
```
cole-h, line 294: This will only log an error, not cause the function to exit with an error; you'll need to `return Err(.........);` after this.

infinisil: I tried this, but I'm not sure where the error should be handled. Ideally already at ofborg startup.

wolfgangwalther:
This was done in 50a2f167f2d7f57d60b1190e574f4c38820e3fb2.

PR #675, opened 2024-02-01
report build failures as failures

adamcstephens:
It is far too easy to miss clear failures when we report them as neutral. Not only does this not register them in GitHub visually as a failure, but once all checks have completed successfully GitHub collapses the checks UI and shows the PR as fully successfully checked.

We're already merging PRs that are failing checks, it's just not as visible. This change will at least provide visibility into those failing checks, to allow us to make better informed decisions when proceeding to merge anyway.

Note: I have little knowledge of what I'm doing, but this seems to be the correct place to change this.

### DIFF
--- a/ofborg/src/message/buildresult.rs
+++ b/ofborg/src/message/buildresult.rs
@@ -30,10 +30,10 @@ impl From<BuildStatus> for Conclusion {
         match status {
             BuildStatus::Skipped => Conclusion::Skipped,
             BuildStatus::Success => Conclusion::Success,
-            BuildStatus::Failure => Conclusion::Neutral,
+            BuildStatus::Failure => Conclusion::Failure,
             BuildStatus::HashMismatch => Conclusion::Failure,
             BuildStatus::TimedOut => Conclusion::Neutral,
-            BuildStatus::UnexpectedError { .. } => Conclusion::Neutral,
+            BuildStatus::UnexpectedError { .. } => Conclusion::Failure,
         }
     }
 }
--- a/ofborg/src/tasks/githubcommentposter.rs
+++ b/ofborg/src/tasks/githubcommentposter.rs
@@ -384,7 +384,7 @@ patching script interpreter paths in /nix/store/pcja75y9isdvgz5i00pkrpif9rxzxc29
                 started_at: None,
                 completed_at: Some("2023-04-20T13:37:42Z".to_string()),
                 status: Some(CheckRunState::Completed),
-                conclusion: Some(Conclusion::Neutral),
+                conclusion: Some(Conclusion::Failure),
                 details_url: Some(
                     "https://logs.ofborg.org/?key=nixos/nixpkgs.2345&attempt_id=neatattemptid"
                         .to_string()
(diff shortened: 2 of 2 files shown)


alyssais:
I don't think timed out should be a failure.  That's something that happens extremely often (basically any PR targeting staging will time out on Darwin), and getting people used to merging despite a reported failure will cause [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue) and make it more likely that real failures are ignored.

ofborg/src/message/buildresult.rs
```diff
@@ -30,10 +30,10 @@ impl From<BuildStatus> for Conclusion {
         match status {
             BuildStatus::Skipped => Conclusion::Skipped,
             BuildStatus::Success => Conclusion::Success,
-            BuildStatus::Failure => Conclusion::Neutral,
+            BuildStatus::Failure => Conclusion::Failure,
             BuildStatus::HashMismatch => Conclusion::Failure,
-            BuildStatus::TimedOut => Conclusion::Neutral,
-            BuildStatus::UnexpectedError { .. } => Conclusion::Neutral,
+            BuildStatus::TimedOut => Conclusion::Failure,
```
mweinelt: Opposed to making timeouts a failure. In most cases there is just too much to build for ofBorg to complete it within the allotted time frame.

alyssais (commented):
Can we distinguish failures of the leaf OfBorg was trying to build from failures of dependencies?  Another thing that comes up a lot especially on staging is failures of dependencies due to flaky tests or whatever, and those shouldn't block a PR being merged, so they shouldn't show up as failures.

ofborg/src/message/buildresult.rs
adamcstephens: Should they stay neutral or would skipped be better?

adamcstephens:
> Can we distinguish failures of the leaf OfBorg was trying to build from failures of dependencies? Another thing that comes up a lot especially on staging is failures of dependencies due to flaky tests or whatever, and those shouldn't block a PR being merged, so they shouldn't show up as failures.

I can't answer the first question as I don't know enough about the entire system. I don't think PRs are blocked regardless on failing ofborg checks.

ofborg/src/message/buildresult.rs
mweinelt: Neutral is probably better, since they were not skipped.

alyssais:
> I don't think PRs are blocked regardless on failing ofborg checks.

Not on a technical level, but people get very nervous about ignoring a failure when merging, and in practice will very rarely do it without being very sure it's okay. That's a useful property that we should preserve, and if we lose it by accident, we won't get a second chance.

hraban:
See related discussion here: https://discourse.nixos.org/t/can-we-make-ofborg-show-broken-tests-as-red/10717/2

adamcstephens:
Happy to reopen at any time.

hraban:
Any reason you closed it, @adamcstephens ? I thought this was a worthwile idea and I hope someone will pick it up someday!

PR #680, opened 2024-05-08
update nixos wiki links

jopejoe1:
We have an official wiki, no reason not to use the links from it instead of the unofficial one.


cole-h:
This isn't even used anymore -- it was from the olden days, back when we were on IRC (and even then, it hadn't been used as the "source of truth" since before I joined the project in ~2020 lol).

Could I ask you to instead just delete the file? Thanks!

PR #688, opened 2024-07-30
fix clippy lints and other minor improvements

abysssol:
I was planning on creating a pull request to improve ofborg's behavior when requesting reviews, but noticed that clippy was emitting a lot of warnings and an error. So, I decided to start by fixing them, and along the way ended up adding other assorted improvements as well.

The project appears rather outdated and unidiomatic in general. It seems a partial rewrite may be in order. Most of the dependencies should probably be updated or replaced ([`traitobject`](https://lib.rs/crates/traitobject) isn't sound since it assumes the layout of fat pointers, but won't be fixed since it's unmaintained). Would there be any problems with replacing [`hubcaps`](https://lib.rs/crates/hubcaps) with [`octocrab`](https://lib.rs/crates/octocrab)?

Notable commits:
07c2ccf1123c842267d67cb3f6fb16030663e33b
9925e497a77f9c55d67a4750e3d715d6ac0c1b7e
5299de378ef8e4546ae1a2954ec3e32210f9678c

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -38,7 +38,7 @@ fn label_from_title(title: &str) -> Vec<String> {
     let labels: Vec<_> = TITLE_LABELS
         .iter()
         .filter(|(word, _label)| {
-            let re = Regex::new(&format!("\\b{}\\b", word)).unwrap();
+            let re = Regex::new(&format!("\\b{word}\\b")).unwrap();
             re.is_match(title)
         })
         .map(|(_word, label)| (*label).into())
@@ -262,8 +262,8 @@ impl<'a> NixpkgsStrategy<'a> {
                 "Potential Maintainers",
                 Some("".to_owned()),
                 match m {
-                    Ok(ref maintainers) => format!("Maintainers:\n{}", maintainers),
-                    Err(ref e) => format!("Ignorable calculation error:\n{:?}", e),
+                    Ok(ref maintainers) => format!("Maintainers:\n{maintainers}"),
+                    Err(ref err) => format!("Ignorable calculation error:\n{err:?}"),
                 },
             );
 
@@ -277,7 +277,7 @@ impl<'a> NixpkgsStrategy<'a> {
                 let status = CommitStatus::new(
                     self.repo.statuses(),
                     self.job.pr.head_sha.clone(),
-                    format!("{}-eval-check-maintainers", prefix),
+                    format!("{prefix}-eval-check-maintainers"),
                     String::from("large change, skipping automatic review requests"),
                     gist_url,
                 );
@@ -288,7 +288,7 @@ impl<'a> NixpkgsStrategy<'a> {
             let status = CommitStatus::new(
                 self.repo.statuses(),
                 self.job.pr.head_sha.clone(),
-                format!("{}-eval-check-maintainers", prefix),
+                format!("{prefix}-eval-check-maintainers"),
                 String::from("matching changed paths to changed attrs..."),
                 gist_url,
             );
@@ -317,7 +317,7 @@ impl<'a> NixpkgsStrategy<'a> {
             let mut status = CommitStatus::new(
                 self.repo.statuses(),
                 self.job.pr.head_sha.clone(),
-                format!("{}-eval-check-meta", prefix),
+                format!("{prefix}-eval-check-meta"),
                 String::from("config.nix: checkMeta = true"),
                 None,
             );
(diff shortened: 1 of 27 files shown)


abysssol:
@cole-h It appears from merged pull requests that you have been the primary ofborg maintainer for some time. Is that still the case?
If so, I would appreciate a review, or at least a timeline on when I could expect a review.
If not, could you point me toward someone who could give a review? I'm not familiar with how pull requests are handled for ofborg.

ofborg/test-srcs/maintainers/lib/attrsets.nix
cole-h, line 1: Everything under test-srcs was copied from Nixpkgs (at one point), please don't change anything in there (mostly because any changes will be just noise).

ofborg/src/tasks/eval/nixpkgs.rs
```diff
@@ -628,16 +630,23 @@ fn parse_commit_messages(messages: &[String]) -> Vec<String> {
...
-
     use super::*;
 
+    trait PipeSort<T: Ord>: Sized + AsMut<[T]> {
+        fn sorted(mut self) -> Self {
+            self.as_mut().sort();
+            self
+        }
+    }
+    impl<T: Ord, L: Sized + AsMut<[T]>> PipeSort<T> for L {}
```
cole-h, line 648: Why's this necessary?

cole-h (commented):
> The project appears rather outdated and unidiomatic in general

I'd agree with that :laughing: 

> Would there be any problems with replacing [hubcaps](https://lib.rs/crates/hubcaps) with [octocrab](https://lib.rs/crates/octocrab)?

In principle, no, but when I played with that in the past (which was a while ago to be fair), I think it was missing something that would severely handicap ofborg's functionality (not that I can remember what that was now, though... maybe checkruns are unsupported?).

ofborg/src/tasks/eval/nixpkgs.rs
abysssol: It's not necessary, it's just there to allow `list.sorted()` method syntax instead of `sorted(list)` function syntax. If you would prefer a bare function without the trait boilerplate, I can easily switch it.

cole-h: Well really my confusion stems from the fact that nothing really seems to have changed... Why does the current `.sort()` call not do what we want here? (Is this something that happened because of a Rust version/edition change?)

abysssol: In short, the [`sort()`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.sort) method does **not** return a list, sorted or otherwise.
The method takes a mutable reference to a slice, sorts the slice in place (mutating the memory), then (crucially) returns [`unit`](https://doc.rust-lang.org/stable/std/primitive.unit.html)/`()`  (similar to null in some other languages).

You may refer to the [std library's documentation of the sort method](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.sort) to confirm this.

> Is this something that happened because of a Rust version/edition change?

No, the method has always worked like this.

---

I did explain this in the relevant commit description. Are you reviewing the commits individually?

If you're not, I would recommend going to the first commit (which is relevant to this change), reviewing there, then using the `previous` and `next` buttons to move between commits to review them one at a time.

I did try to organize the commits by their purpose and leave a commit description on more significant changes for easier review.

cole-h: Ah, right, I always forget that `.sort()` does its thing in-place and doesn't return the slice. Oops.

> Are you reviewing the commits individually?

When there are more than a handful of commits, I usually review the entire diff at once, rather than commit-by-commit. Mostly because I've gotten used to "lots of commits" being usually not well-focused (I'm guilty of this myself, on occasion).

Now that I've gone back through commit-by-commit, your explanation (here and in the commit) makes sense, thanks :)

cole-h (approved):
Diff LGTM.

I'll attempt to deploy this sometime this week -- if all goes smoothly, I'll come back and merge it :rocket: 

Thanks!!

abysssol:
By the way, now that you're done with review, would you like me to squash some of the commits to reduce the quantity of minor commits? I could squash some of the minor clippy fixes, and maybe a couple others. I don't know if it matters, though. Are there any guidelines on commit quantity and size? Any personal preferences?

Do  you have any general feedback on how I structured the commits? Anything I could improve?

Thank you for all your help.

cole-h:
> would you like me to squash some of the commits to reduce the quantity of minor commits

No need to squash, I'll just rebase-merge as-is.

> Are there any guidelines on commit quantity and size? Any personal preferences?
>
> Do you have any general feedback on how I structured the commits? Anything I could improve?

The commits were good, no specific guidance there.

---

Been deployed for ~1.5h now and seems to be working fine, so I'm gonna merge it now. Thanks again for your effort!

PR #706, opened 2025-05-26
Small improvements to check runs

### LINKED ISSUE #283: Checks UI, tab labels should be different
image

The tab labels are all identical after truncation, which makes them useless. They could instead display the difference, here `x86_64-darwin` vs `aarch64-linux`.

wolfgangwalther:
With long status entries with many packages, the system is cut off. Since the system is fixed length, it makes more sense to put it at the front. This allows to easily identify the relevant checks.

Also avoid posting a status check for everything with "/" in the prefix, because that's not a package to build anyway.

Resolves #283

### DIFF
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -174,6 +174,10 @@ fn parse_commit_messages(messages: &[String]) -> Vec<String> {
             // Convert "foo: some notes" in to "foo"
             line.split_once(':').map(|(pre, _)| pre.trim())
         })
+        .filter(|line| {
+            // When a prefix contains a slash, it targets a certain file, but not a buildable package.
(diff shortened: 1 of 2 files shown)


wolfgangwalther:
@dasJ maybe you can have a look at this one?

wolfgangwalther:
One slight annoyance with the change: Previously all 4 ofborg jobs would appear next to each other, because they all start with the attrpath. After this change, they would be separate, because two start with aarch64 and two start with x86_64 - thus appearing at opposite ends of the check list. I added the shortest possible common prefix to all of them: A `.`. This should make them all end up in the same position.

wolfgangwalther:
@Conni2461 could you look at this, too?

blexzy:
Good

wolfgangwalther:
There seems to be no interest in improving this.

Issue #695, opened 2024-11-16
Future development of Ofborg


Mic92:
Since there is the question what we should do this in future, it would be good to figure out,
who would like to work on this and how much time you think you could spend on this.

Contribution could fall into one of the two categories or both:

1. Infrastructure maintenance
2. Development

crertel:
I'm interested in helping on this, though my knowledge is not great. It feels like there's maybe a three-stage problem here:

1. Stage 1: How do we secure some breathing room so people aren't working through Christmas or making a botnet to exploit Github Actions?
2. Stage 2: How do we moved to cheaper hosting/secure hardware of our own and park it somewhere?
3. Stage 3: What options to we have to reduce the footprint of the project entirely so that we don't have have to keep relying on donations/corporate sponsorship?

Mic92:
>     1. Stage 1: How do we secure some breathing room so people aren't working through Christmas or making a botnet to exploit Github Actions?

At the moment it looks like there might be no botnet needed (and also I haven't seen anyone suggesting it). Infinisil managed to evaluate all 4 architectures with 4 Actions in under 5 minutes while also using `./pkgs/top-level/release-outpaths.nix` (https://github.com/NixOS/nixpkgs/pull/356023/files). On github we have concurrent 60 runners, so we should have some basic coverage for a while and hopefully get faster results than what ofborg currently delivers where the queue time is currently 18h (?) for evaluation.

crertel:
That's really great news! Good work @infinisil !

EDIT: also thank you amjoseph is appears!

Mic92:
amjoseph did a good part of this work. It was actually supposed to end up in ofborg before he took a break from the nix-community.

eveeifyeve:
> 2. Stage 2: How do we moved to cheaper hosting/secure hardware of our own and park it somewhere?

What would be a cheeper infra then github actions, would be what bun currently uses which is buildkite which has a much cheeper pricing then github actions.

wolfgangwalther:
I would like to help and go through all the issues and PRs to close those that don't matter anymore, now that we do a lot of stuff in nixpkgs. Would be great, if I could get triaging rights here, to be able to close issues and PRs.

infinisil:
Only @cole-h is admin of this repo, but it doesn't look like they're active anymore.

@NixOS/org I propose making @wolfgangwalther and @dasJ (who is the most recently active person working on this repo) admins of this repo, sound good?

cole-h:
[...]

PR #708, opened 2025-07-07
Remove merge conflict label

wolfgangwalther:
This label is now managed by Nixpkgs CI. Technically it doesn't hurt to manage it via ofborg, too, but ofborg is *too quick* sometimes. The scheduled labeler in nixpkgs CI works with a bit of delay and thus doesn't temporarily label a PR conflicted while changing base branches and pushing again.

As discussed in https://github.com/NixOS/nixpkgs/pull/422906#issuecomment-3041481371.

cc @philiptaron

### DIFF
--- a/ofborg/src/tasks/eval/mod.rs
+++ b/ofborg/src/tasks/eval/mod.rs
@@ -13,7 +13,6 @@ pub trait EvaluationStrategy {
 
     fn on_target_branch(&mut self, co: &Path, status: &mut CommitStatus) -> StepResult<()>;
     fn after_fetch(&mut self, co: &CachedProjectCo) -> StepResult<()>;
-    fn merge_conflict(&mut self);
     fn after_merge(&mut self, status: &mut CommitStatus) -> StepResult<()>;
     fn evaluation_checks(&self) -> Vec<EvalChecker>;
     fn all_evaluations_passed(
--- a/ofborg/src/tasks/eval/nixpkgs.rs
+++ b/ofborg/src/tasks/eval/nixpkgs.rs
@@ -128,21 +128,7 @@ impl<'a> EvaluationStrategy for NixpkgsStrategy<'a> {
         Ok(())
     }
 
-    fn merge_conflict(&mut self) {
-        update_labels(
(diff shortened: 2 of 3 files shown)


philiptaron (approved):
Yep. I'm not familiar with the ofBorg branches: is `released` the correct target?

wolfgangwalther:
I am not either, but it's the *only* branch (except two dependabot branches), so.. yes :)

wolfgangwalther:
In [424162](https://github.com/NixOS/nixpkgs/pull/424162) ofborg and nixpkgs-ci are playing ping-pong with the merge conflict label. Clearly, ofborg has other ideas about the merge status than nixpkgs-ci and GitHub.

So we should really disable this.

@dasJ could you please look at this PR or alternatively onboard me in terms of how to merge and deploy stuff as mentioned in https://github.com/NixOS/ofborg/issues/695#issuecomment-2949763443?

sternenseemann:
@wolfgangwalther ofborg is right in that case as there is no actual conflict, i.e. git can merge that branch automatically.

wolfgangwalther:
Well, "being right" is not what we want to indicate, though. GitHub's UI doesn't allow merging, because it identifies a merge conflict. The label should reflect that, to be able to easily filter out PRs where you can't hit the button.

Conni2461:
diff looks good to me, thanks. i cant merge here, i asked dasJ in private if he can merge, but i just deployed that change to production: https://github.com/ofborg/ofborg-infrastructure/commit/18819813d2a25a872daf45e281773a38ac51b9c8

Conni2461 (approved):

wolfgangwalther:
@Conni2461 is that deployment live already?

In https://github.com/NixOS/nixpkgs/pull/424162, ofborg is still doing something about this label. Did I forget any related code? I don't think so.

Conni2461:
you were right i havent deployed the right host the first time i deployed, but i fixed that right after you pinged me last week and i just checked the logs and ofborg didnt add another label for merge conflicts since then. so this is good to go.

thanks again :)

PR #713, opened 2025-09-02
fix: hubscups, bump to fork to fix parsing errors

Conni2461:


wolfgangwalther:
Oh my... wow.

> Allow larger ids for CheckSuite.id deserialization

These IDs hit the 32bit limit? :D

wolfgangwalther (approved):

Conni2461:
> These IDs hit the 32bit limit? :D

that issue was an issue for way to long, but it doesnt matter because thats the response and at this point in time the item is already created so its just a lot of spam in the logs so i fixed that one as well.

i explained the issue in the ofborg matrix channel:

> for anyone interested: issue was that https://api.github.com/repos/nixos/nixpkgs/commits/{sha}/statuses returned a response which didnt parse sometimes. hubcaps expected that description to always be String, but github was returning null for some. example: curl https://api.github.com/repos/nixos/nixpkgs/commits/efd958b34fa5e157368d61c3401546e2c0bd0d15/statuses

wolfgangwalther:
Uh, that would have been... me, then: https://github.com/NixOS/nixpkgs/pull/435929/commits/2c25cb0891d23f052baac7db22004ec39e683921

This commit posts a new status without description. Who would have thought that I'd break ofborg with that :D

Thanks for fixing!

PR #715, opened 2025-12-11
Bump manually deps, async-std -> tokio

Ericson2314:
Last commit:

> The LLM did this, and I cannot (before or after) run all the tests without things hanging, so we should procede with caution.

The other commits are simple enough that they are easy to hand-review.


Conni2461:
changes were included in https://github.com/NixOS/ofborg/pull/718

PR #718, opened 2026-01-16
Bump manually deps, async-std -> tokio

Conni2461:
includes https://github.com/NixOS/ofborg/pull/715 and then replaces block_on with async functions everywhere.

this is already deployed and tested, everything is working smoothly so far


Conni2461:
i will most likely merge this tomorrow, if there arent any regressions till then.

PR #728, opened 2026-04-17
fix: Don't return info on dirs outside of logs_path

Erethon:
This PR fixes a minor security issue that leaks filenames under specific conditions. If you can access the logapi endpoint directly or if a reverse proxy isn't canonicalizing paths in requests, it's possible to leak filenames in any directory the ofborg user can access that doesn't contain other subdirectories.

Steps to reproduce (in the ofborg repo):
```
cat > conf << 'EOF'
{
  "log_api_config": {
    "listen": "0.0.0.0:8080",
    "logs_path": "/var/lib/"
  },
  "runner": {
      "identity": "logapi",
      "repos": [],
      "disable_trusted_users": true,
      "trusted_users": []
  },
  "checkout": {
      "root": "/tmp/ofborg-checkout"
  },
  "nix": {
      "system": "x86_64-linux",
      "remote": "daemon",
      "build_timeout_seconds": 3600,
      "initial_heap_size": "4g"
  }
}
EOF

nix-shell
cargo run --bin logapi -- conf

curl --path-as-is 127.0.0.1:8080/logs/../../etc/ssh
{"attempts":{"ssh_host_rsa_key":{"metadata":null,"result":null,"log_url":"https://logs.ofborg.org/logfile/../../etc/ssh/ssh_host_rsa_key"},"ssh_host_ed25519_key.pub":{"metadata":null,"result":null,"log_url":"https://logs.ofborg.org/logfile/../../etc/ssh/ssh_host_ed25519_key.pub"},"initrd_ssh_host_ed25519_key":{"metadata":null,"result":null,"log_url":"https://logs.ofborg.org/logfile/../../etc/ssh/initrd_ssh_host_ed25519_key"},"ssh_host_ed25519_key":{"metadata":null,"result":null,"log_url":"https://logs.ofborg.org/logfile/../../etc/ssh/ssh_host_ed25519_key"},"ssh_host_rsa_key.pub":{"metadata":null,"result":null,"log_url":"https://logs.ofborg.org/logfile/../../etc/ssh/ssh_host_rsa_key.pub"}}}
```

I've also added a check that `logs_path` exists since I ran into an issue with it while experimenting with this bug.

Edit: I've already synced with @mweinelt about opening this PR since it's not important enough to warrant an embargo.

### DIFF
--- a/ofborg/src/bin/logapi.rs
+++ b/ofborg/src/bin/logapi.rs
@@ -26,7 +26,7 @@ struct LogResponse {
 
 #[derive(Clone)]
 struct LogApiConfig {
-    logs_path: String,
+    logs_path: PathBuf,
     serve_root: String,
 }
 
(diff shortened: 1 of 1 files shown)


Conni2461:
thanks :)

